PR #22, opened 2016-12-14
Added option for curvilinear grid, including rotation angles of grid

sambarluc:
I added an option to separately deal with Cartesian curvilinear grids. The only difference with standard Cartesian is that sine/cosine defining the grid rotation at each point are now being read.

### DIFF
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -102,7 +102,19 @@ _experiments = {
                             'SIatmFW', 'oceQnet', 'oceFWflx', 'oceTAUX',
                             'oceTAUY', 'ADVxHEFF', 'ADVyHEFF', 'DFxEHEFF',
                             'DFyEHEFF', 'ADVxSNOW', 'ADVySNOW', 'DFxESNOW',
-                            'DFyESNOW', 'SIuice', 'SIvice'])}
+                            'DFyESNOW', 'SIuice', 'SIvice'])},
+    'curvilinear_leman': {'geometry': 'curvilinear',
+                          'delta_t': 20,
+                          'ref_date': "2013-11-12 12:00",
+                          'shape': (35, 64, 340),
+                          'test_iternum': 6,
+                          'dtype': np.dtype('f4'),
+                          'expected_values': {'XC': ((0,0), 501919.0)},
+                          'all_iters': [0, 3, 6],
+                          'expected_time':[
+                            (0, np.datetime64('2013-11-12T12:00:00.000000000')),
+                            (1, np.datetime64('2013-11-12T12:02:00.000000000'))],
+                          'prefixes': ['THETA']}
 }
 
 
@@ -423,7 +435,7 @@ def test_swap_dims(all_mds_datadirs):
     # make sure we never swap if not reading grid
     assert 'i' in xmitgcm.open_mdsdataset(dirname,
         iters=None, read_grid=False, geometry=expected['geometry'])
-    if expected['geometry'] == 'llc':
+    if expected['geometry'] in ('llc', 'curvilinear'):
         # make sure swapping is not the default
         assert 'i' in xmitgcm.open_mdsdataset(dirname, **kwargs)
         # and is impossible
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -109,7 +109,7 @@ _experiments = {
                           'shape': (35, 64, 340),
                           'test_iternum': 6,
                           'dtype': np.dtype('f4'),
-                          'expected_values': {'XC': ((0,0), 501919.0)},
+                          'expected_values': {'XC': ((0,0), 501919.21875)},
                           'all_iters': [0, 3, 6],
                           'expected_time':[
                             (0, np.datetime64('2013-11-12T12:00:00.000000000')),
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -75,6 +75,27 @@ horizontal_coordinates_spherical = OrderedDict(
                 units="degrees_north", coordinates="YG XG"))
 )
 
+horizontal_coordinates_curvcart = OrderedDict(
+    XC = dict(dims=["j", "i"], attrs=dict(
+                standard_name="plane_x_coordinate", long_name="x coordinate",
+                units="m", coordinate="YC XC")),
+    YC = dict(dims=["j", "i"], attrs=dict(
+                standard_name="plane_y_coordinate", long_name="y coordinate",
+                units="m", coordinate="YC XC")),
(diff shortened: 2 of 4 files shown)


xmitgcm/variables.py
```diff
@@ -75,6 +75,27 @@
...
+                long_name="x coordinate", units="m", coordinate="YG XG")),
+    YG = dict(dims=["j_g", "i_g"], attrs=dict(
+                standard_name="plane_y_coordinate_at_f_location",
+                long_name="y coordinate", units="m", coordinates="YG XG")),
+    CS  = dict(dims=["j", "i"], attrs=dict(standard_name="Cos of grid orientation angle",
+               long_name="AngleCS", units=" ", coordinate="YC XC"),
+               filename='AngleCS'),
+    SN  = dict(dims=["j", "i"], attrs=dict(standard_name="Sin of grid orientation angle",
+               long_name="AngleSN", units=" ", coordinate="YC XC"),
+               filename='AngleSN')
```
rabernat: Is there a reason to call these variables `SN` and `CS` rather than just `AngleSN` and `AngleCS`?

In the only cases where I change the grid names from their filename, I did it to match what the MITgcm mnc output does.

sambarluc: `SN` and `CS` are the netcdf names, yes, I tried to follow what you had done for the other variables.

Concerning the test, I noticed the complaint in the pull request, but there are no verifications available using curvilinear grid but not LLC (happy to be proven wrong). I will try to make a small test case based on my configuration but I am afraid that will take a little time.

rabernat: The test example for the llc geometry is described here:
https://github.com/xgcm/xmitgcm/blob/master/xmitgcm/test/test_mds_store.py#L90
(The files are stored in xmitgcm/test/global_oce_llc90.tar.gz)

I call the llc geometry just `llc`, but perhaps curvilinear would be a better description. Maybe `llc` means something else, which is related to exch2. In that case, we already have a test case we can use.

rabernat:
Ping! Any progress on this @sambarluc?

sambarluc:
I'm afraid, not at all, and won't make any before next year, I'm very sorry. A very busy end of year between flu, proposals and much needed holidays.

rabernat:
I think we should probably replace `geometry='llc'` with `geometry='curvilinear'`, since the llc geometry is in fact curvilinear. Since llc is not actually a geometry, perhaps we need a separate option for llc, like simply `llc=True` (which would automatically imply `geometry='curvilinear').

sambarluc:
Hi there, I am finally working on a test for the curvilinear grid. I am struggling with a last failure in `pytest`. In `test_values_and_endianness[curvilinear_leman]` I get the assertion error:
```
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           (mismatch 100.0%)
E            x: array(501919.21875, dtype=float32)
E            y: array(501919.0)

test_mds_store.py:376: AssertionError
```
The value I believe is the correct one is 501919.0. How should I interpret `x` and `y` arrays? This should be the check in the endianness, is that 0.22 some sort of rounding error?
(sorry, I know I am very ignorant on these things, please do laugh if I'm saying bullshit)

Thanks, Andrea

rabernat:
This is a strange error. Could you just push your current branch so that we can see your full code? It doesn't have to be finished.

You might have to rebase because there have been lots of changes to master recently (e.g. #23, #32).

sambarluc:
This should be it. At least on my machine, one failure as described above.

xmitgcm/test/test_mds_store.py
```diff
@@ -102,7 +102,19 @@ def hide_file(origdir, *basenames):
...
                             'DFyEHEFF', 'ADVxSNOW', 'ADVySNOW', 'DFxESNOW',
-                            'DFyESNOW', 'SIuice', 'SIvice'])}
+                            'DFyESNOW', 'SIuice', 'SIvice'])},
+    'curvilinear_leman': {'geometry': 'curvilinear',
+                          'delta_t': 20,
+                          'ref_date': "2013-11-12 12:00",
+                          'shape': (35, 64, 340),
+                          'test_iternum': 6,
+                          'dtype': np.dtype('f4'),
+                          'expected_values': {'XC': ((0,0), 501919.0)},
```
rabernat: This is where the value is set.

rabernat:
Ok @sambarluc thanks for this.

Yes, I can see that Travis CI is getting the same error. The error arises due to a mismatch between the value that xmitgcm reads from the file 'XC.data' and the expected value, which you specified to be 501919.0 in the test metadata.

To check this out, I downloaded `curvilinear_lmean.tar.gz` from your branch and manually examined the file:

```python
>>> data = np.fromfile('XC.data', dtype='>f4')
>>> print('%10f' % data[0])
501919.218750
```

So it looks like your expected value is actually incorrect.

Curiously, there seems to be some weird behavior involving print. If I just call print with no formatting,
```python
>>> print(data[0])
501919.0
```
I get the number with no decimals. So this might be the reason you put in that expected value. That is really strange and feels like a bug to me. But there is no doubt that the correct value is 501919.21875.

It seems to print correctly with a slice
```python
>>> print(data[:1])
[ 501919.21875]
```

xmitgcm/variables.py
rabernat: Following up on our earlier discussion, it seems like we should use this pull request to add reading of `CS` and `SN` for the LLC geometry as well.

sambarluc:
Ok, I fixed this and now it completes the tests successfully. I was tricked by the issue you mention. I really don't get it but at least it's fixed.

rabernat:
I wanted to get your opinion on the relationship between curvilinear and llc. If you look at the [llc data files](http://wwwcvs.mitgcm.org/viewvc/MITgcm/MITgcm_contrib/llc_hires/llc_1080/input/data?revision=1.9&view=markup#l98), you see that `usingCurvilinearGrid=.TRUE.`. However, it is a curvilinear grid on the sphere, while yours is cartesian.

Do you think it makes sense to just allow `geometry={'cartesian', 'sphericalpolar', 'curvilinear'}` and make `llc` into a separate parameter (related to the presence of multiple facets)? And if so, how do we distinguish from cartesian vs spherical curvilinear?

sambarluc:
I am not familiar with "llc" so the opinion is not based on direct experience. Personally, I see no added value in reproducing the MITgcm grid hierarchy, while having two separate option "llc" and "curvilinear" is in my view easier for the user.

rabernat:
Ok, let's leave it as is for now.

One final request before merging: could you edit the documentation to describe the new option for grid geometry?
https://github.com/xgcm/xmitgcm/blob/master/doc/index.rst#grid-geometries

rabernat:
Great! Thanks so much for this contribution

sambarluc:
Thank you, I'm finding xmitgcm very useful in my work!

PR #23, opened 2016-12-15
Added fix_inconsistent_variables

braaannigan:
I added the function fix_inconsistent_variables to add the extra row onto drC.  This function is called after the data is read from the mds files but before the data is read into the xarray dataset.  Trying to change the drC array once it is in the xarray dataset causes problems, because xarray generates an error because the shape is different to what it expects.

I also changed read_grid=False to True on Line 165, as this seems to override the choice made when the user calls the function and so the grid wasn't being read.  I'm not sure what this should be in general.

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -153,7 +153,7 @@ def open_mdsdataset(data_dir, grid_dir=None,
                     nx=nx, ny=ny, nz=nz)
                 datasets = [open_mdsdataset(
 
-                        data_dir, iters=iternum, read_grid=False, **kwargs)
+                        data_dir, iters=iternum, read_grid=True, **kwargs)
                     for iternum in iters]
                 # now add the grid
                 if read_grid:
@@ -408,18 +408,34 @@ class _MDSDataStore(xr.backends.common.AbstractDataStore):
         prefixes = []
         if read_grid:
             prefixes = prefixes + list(self._all_grid_variables.keys())
-
+            
         # add data files
         prefixes = (prefixes +
                     _get_all_matching_prefixes(
                                                data_dir,
                                                iternum,
                                                file_prefixes))
-
+    #In some older versions of mitgcm len(drC)=nr rather than nr+1
+    #Define a function to change it to the newer format if len(drC)=nr
+    #and apply it to the data before it is added to the xarray variable below
+    def fix_inconsistent_variables(data):
+        # check if the drC variable has the wrong length
+        if len(data)==self.nz:
+            # create a new array which will replace it
+            drc_data = np.zeros(self.nz + 1)
+            drc_data[:-1] = data
+            # fill in the missing value
+            drc_data[-1] = 0.5 * data[-1]
+            data = drc_data
+        return data
+    
         for p in prefixes:
             # use a generator to loop through the variables in each file
             for (vname, dims, data, attrs) in self.load_from_prefix(p, iternum):
                 # print(vname, dims, data.shape)
+                if vname == 'drC':
+                    data = fix_inconsistent_variables(data)
+
                 thisvar = xr.Variable(dims, data, attrs)
                 self._variables[vname] = thisvar
                 # print(type(data), type(thisvar._data), thisvar._in_memory)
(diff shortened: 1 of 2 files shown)


braaannigan:
The indent on the new fix_inconsistent_variables was wrong in my first commit, I've modified this now

braaannigan:
It's working now for both the test case and my own experiment now.  It's great!

xmitgcm/mds_store.py
```diff
@@ -153,7 +153,7 @@ def open_mdsdataset(data_dir, grid_dir=None,
                     nx=nx, ny=ny, nz=nz)
                 datasets = [open_mdsdataset(
 
-                        data_dir, iters=iternum, read_grid=False, **kwargs)
+                        data_dir, iters=iternum, read_grid=True, **kwargs)
```
rabernat: Could you change this back? If you look more carefully at the code path, you will see that this is just part of the logic for how the dataset is put together. If `read_grid` is passed as `True` to `open_mdsdataset`, the function calls itself recursively. There are extensive tests ([example](https://github.com/xgcm/xmitgcm/blob/master/xmitgcm/test/test_mds_store.py#L335) to ensure that the options work properly. I am surprised this change didn't cause an error somewhere.

xmitgcm/mds_store.py
```diff
@@ -408,18 +408,35 @@ def __init__(self, data_dir, grid_dir=None,
...
                     _get_all_matching_prefixes(
                                                data_dir,
                                                iternum,
                                                file_prefixes))
-
+    
+        #In some older versions of mitgcm len(drC)=nr rather than nr+1
+        #Define a function to change it to the newer format if len(drC)=nr
+        #and apply it to the data before it is added to the xarray variable below
+        def fix_inconsistent_variables(data):
```
rabernat: Do you think you could make this function have the signature?
```python
vname, dims, data, attrs = self._fix_inconsistent_variables(vname, dims, data, attrs)
```
Then you could do the check ``if vname=='drC'`` inside that function. (If false, you just return `vname, dims, data, attrs` with no modifications.) I would prefer this because it will keep the code path within ``__init__`` much cleaner, and it will also generalize in case we have to do more such hacking on other variables.

rabernat: It should also be moved outside of ``__init__`` and become its own method of the  `_MDSDataStore` class.

braaannigan:
Made those changes, it's still working for me

rabernat:
@braaannigan, just curious, do you know how to run the test suite yourself?

```bash
py.test -v xmitgcm
```

Can be useful for development.

braaannigan:
Hi @rabernat, this is all new to me!

In test/test_mds_store.py, what exactly needs to be added?  Is it adding a new experiment to the dictionary on L60 with similar properties to global_oce_latlon?

Also, I'm going hiking for a few days, so I'll probably be offline from about 8am tomorrow until Tuesday.  Hopefully we can get this sorted before then

rabernat:
> Hi @rabernat, this is all new to me!

Yes, it was new to me a few years ago too. Once you start testing, you will never go back. 😉 

Add a function to test_mds_store.py that goes something like this
```python
def test_drc_length(all_mds_datadirs):
    dirname, _ = all_mds_datadirs
    copyfile(os.path.join(dirname, 'DRF.data'), os.path.join(dirname, 'DRC.data'))
    copyfile(os.path.join(dirname, 'DRF.meta'), os.path.join(dirname, 'DRC.meta'))
    ds = xmitgcm.open_mdsdataset(
                dirname, iters=None, read_grid=True,
                geometry=expected['geometry'])
    assert len(ds.drC)==(len(ds.drF)+1)
```

You will also have to add ``from shutil import copyfile`` at the top of test_mds_store.py.

braaannigan:
Great, that worked once I added 'expected' to the first line of the
function.  All tests passed.


2016-12-15 20:59 GMT+01:00 Ryan Abernathey <notifications@github.com>:

rabernat:
Awesome! Your test (along with all the others) was run by [travis CI](https://travis-ci.org/xgcm/xmitgcm/builds/184350937) on this pull request, and it passed. It will continue to be run every time someone makes a new pull request. This is how we ensure we don't break things as the package evolves.

You can also see that the [code coverage checks](https://codecov.io/gh/xgcm/xmitgcm/compare/21be4636c7791d79c7dfa117f110a92b901a1bc6...2739dfe93756681c0d142ced6c7453d938b90752) also passed. This means that the new code you added also came with sufficient tests.

braaannigan:
Cheers Ryan, thanks for your help!

PR #25, opened 2017-01-02
Wrap data with dask

rabernat:
This is a pretty major refactor of how the llc data is read. We now construct a fancy dask graph to load the data in face / level chunks.


jklymak:
@rabernat  I missed this.  Does it effectively close https://github.com/xgcm/xgcm/issues/40 ?

rabernat:
> Does it effectively close [xgcm/xgcm#40](https://github.com/xgcm/xgcm/issues/40) ?

No, that issue was resolved years ago, with a major internal refactor of xmitgcm. Why don't you try it out and see if it's any better with your use case.

PR #34, opened 2017-01-30
Slight modifications to improve the use of the default_dtype parameter

serazing:
* Check the default_dtype to add if necessary the default endian

* Modify the IOError when the metadata is absent and the default_dtype is needed

* Add a test to check if an IOError is correctly raised when there is no metadata and default_dtype is None


rabernat:
Simple and perfect solution. And it is covered by tests. Thanks so much for the contribution!

PR #46, opened 2017-06-25
Added new option 'diagnostics' for open_mdsdataset() function

mobigroup:
Added new option 'diagnostics' to use custom version of the default_diagnostics.log. As sample: open_mdsdataset(...,diagnostics='custom_diagnostics.log'). It can be used when we have 2D/3D variables with the same names mixed in one data directory.


mobigroup:
Sorry, it's requires for some checks before.

PR #50, opened 2017-10-10
added KPP state variables to list

dhruvbalwada:
Added the instantaneous state variables from KPP to the list.

Was unable to add the time averaged fields, because the KPP package assigns them names like "KPPhbl-T.data", and Python doesn't allow a variable to be named with a minus sign.  
Any suggestions would be appreciated.


xmitgcm/variables.py
```diff
@@ -275,6 +275,27 @@
...
+               long_name='KPP vertical diffusion coefficient for heat',
+               units='m2 s-1')),
+    KPPghat    = dict(dims=['k_l','j','i'], attrs=dict(
+               standard_name="KPP_non_local_coeff",
+               long_name='non-local transfer coefficient',
+               units='s m-2 ')),
+    KPPhbl   = dict(dims=['j','i'], attrs=dict(
+               standard_name="KPP_boundary_layer",
+               long_name='KPP boundary layer depth, bulk Ri criterion',
+               units='m')),
```
rabernat: I would prefer to have these in a separate data structure, rather than together with state variable. This will also get you around the naming issue, because you can use regular dict syntax.

So at the end of this module, add
```python
kpp_metadata = {'KPPviscAz': (dims=[...
```

We then have to merge these separate dicts into one big one. I will think about how to do that.

rabernat:
It would be great to have these variables in the test datasets so we can verify they work. That is kind of a pain though.

xmitgcm/variables.py
dhruvbalwada: Ok yeah, makes sense. I'll do that.

rabernat: There is currently some logic for concatting dicts and adding the `-T` suffix already in mds_store.py: https://github.com/xgcm/xmitgcm/blob/master/xmitgcm/mds_store.py#L687

Maybe this function will have to be refactored a bit to accomodate more metadata dicts.

The larger issue here is that there are many packages that can potentially do this kind of output (e.g. GMRedi, GGL, EXF, etc.). I hope your PR can create a framework for eventually adding all of these.

rabernat:
Do you need help moving this forward?

PR #53, opened 2017-10-18
Tests for ref_date issue

sambarluc:
8 tests which either open or open+print a dataset in 4 different ways, always using `ref_date`, plus `read_grid=True/False` and `read_grid=True, swap_dims=True/False`

### DIFF
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -749,78 +749,44 @@ def test_drc_length(all_mds_datadirs):
 # Series of tests which try to open a dataset with different combinations of
 # of options, to identify if ref_date can trigger an error
 #
-def test_print_with_ref_date_without_grid(mds_datadirs_with_refdate):
+@pytest.mark.parametrize("method", [True, False])
+def test_print_with_ref_date_grid(mds_datadirs_with_refdate, method):
     """With ref_date, without grid."""
     dirname, expected = mds_datadirs_with_refdate
 
     ds = xmitgcm.open_mdsdataset(dirname, iters='all', prefix=['S'],
-                                 ref_date=expected['ref_date'], read_grid=False,
+                                 ref_date=expected['ref_date'], read_grid=method,
                                  delta_t=expected['delta_t'],
                                  geometry=expected['geometry'])
     print(ds)
 
-def test_print_with_ref_date_with_grid(mds_datadirs_with_refdate):
+@pytest.mark.parametrize("method", [True, False])
+def test_print_with_ref_date_swap(mds_datadirs_with_refdate, method):
     """With ref_date, without grid."""
     dirname, expected = mds_datadirs_with_refdate
 
     ds = xmitgcm.open_mdsdataset(dirname, iters='all', prefix=['S'],
                                  ref_date=expected['ref_date'], read_grid=True,
-                                 delta_t=expected['delta_t'],
+                                 delta_t=expected['delta_t'], swap_dims=method,
                                  geometry=expected['geometry'])
     print(ds)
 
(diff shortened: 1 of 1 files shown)


xmitgcm/test/test_mds_store.py
```diff
@@ -742,3 +743,84 @@ def test_drc_length(all_mds_datadirs):
...
+                                 geometry=expected['geometry'])
+
+def test_open_with_ref_date_without_swap(mds_datadirs_with_refdate):
+    """With ref_date, without grid."""
+    dirname, expected = mds_datadirs_with_refdate
+
+    ds = xmitgcm.open_mdsdataset(dirname, iters='all', prefix=['S'],
+                                 ref_date=expected['ref_date'], read_grid=True,
+                                 delta_t=expected['delta_t'], swap_dims=False,
+                                 geometry=expected['geometry'])
```
rabernat: @sambarluc sorry to be picky, but this is quite repetitive and consequently hard to parse.

Could you parametrize this into one single test?
https://docs.pytest.org/en/latest/parametrize.html

rabernat: See, for example `test_llc_dims`

xmitgcm/test/test_mds_store.py
```diff
@@ -742,3 +743,84 @@ def test_drc_length(all_mds_datadirs):
...
+
+def test_print_with_ref_date_with_swap(mds_datadirs_with_refdate):
+    """With ref_date, without grid."""
+    dirname, expected = mds_datadirs_with_refdate
+
+    ds = xmitgcm.open_mdsdataset(dirname, iters='all', prefix=['S'],
+                                 ref_date=expected['ref_date'], read_grid=True,
+                                 delta_t=expected['delta_t'], swap_dims=True,
+                                 geometry=expected['geometry'])
+    print(ds)
```
rabernat: This is failing with `global_oce_llc90` not due to the time issue but due to the fact that you can't swap dims on llc.

xmitgcm/test/test_mds_store.py
```diff
@@ -742,3 +743,84 @@ def test_drc_length(all_mds_datadirs):
...
+#
+def test_print_with_ref_date_without_grid(mds_datadirs_with_refdate):
+    """With ref_date, without grid."""
+    dirname, expected = mds_datadirs_with_refdate
+
+    ds = xmitgcm.open_mdsdataset(dirname, iters='all', prefix=['S'],
+                                 ref_date=expected['ref_date'], read_grid=False,
+                                 delta_t=expected['delta_t'],
+                                 geometry=expected['geometry'])
+    print(ds)
```
rabernat: Is there a simpler way to trigger the error than calling `print()` on the dataset? What about `ds.load()`. Or even better `ds.time.load()`. I just want to get down to the root of the issue and not mix it with other things.

sambarluc:
It's a nightmare.

sambarluc:
And, btw, my idea that it is linked to python 3 string formatting is clearly wrong, as it fails also on 2.7

rabernat:
Is there a simpler way to trigger the error than calling `print()` on the dataset? What about `ds.load()`. Or even better `ds.time.load()`. I just want to get down to the root of the issue and not mix it with other things.

If we can sort that out, I will merge this and try on my own machine.

sambarluc:
Coming with `ds.time.load()`

rabernat:
Ok I will give it a try.

rabernat:
So if you look carefully at the most recent travis-ci test for python 3.6, you see that the only failure is for swap_dims=True with llc geometry, which is a forbidden option. All of the actual tests did pass with python 3.6.
https://travis-ci.org/xgcm/xmitgcm/jobs/289563761

PR #63, opened 2018-01-14
Added functionality to read PTRtave files

roxyboy:


roxyboy:
I deleted the units in the attributes as this depends on what the user intends the ptracers to be.

xmitgcm/variables.py
```diff
@@ -386,7 +386,14 @@
    state_variables[trname] = dict(dims=['k','j','i'],
        attrs=dict(standard_name='%s_concentration' % trname,
                   long_name="Concentration of %s" % trname,
-                  units="kg m-3")
+                  units="")
```
rabernat: I'm not sure this change is consistent with `available_diagnostics.log`. Can you double check?

roxyboy: I am not quite sure how the log file is generated but is not possible to have a blank for the units? I deleted the units because this depends on the units prescribed in the `data.ptracers` file.

rabernat: I don't want the units to be blank. I want the units to match the tracer units in `available_diagnostics.log`. Once you fix this, I can merge this PR.

roxyboy: @rabernat I can put back the units that was prescribed before I deleted them out but the units are dependent on the units the users chooses to use in `data.ptracers`, e.g. the units were originally hardcoded as "kg m^-3" but this is not the case for my run at least as the Darwin package uses the units of "mmol m^-3".

rabernat: What does `available_diagnostics.log` give as the units of the tracer concentration?

I think `kg m-3` is still better than no units.

roxyboy: @rabernat In my `available_diagnostics.log` the units for `TRAC` are `mmol m-3` so I was initially got confused until I understood that xmitgcm hardcodes the units. But I'll just put the units back in as `kg m-3` for now.

PR #77, opened 2018-03-16
Generalizes _get_all_iternums in order to handle compressed data

apatlpo:
The purpose of this modification is to be able to load iteration numbers when file format that are NOT `*.??????????.data`.
You may for example use `_get_all_iternums` as a utility function in order to get iteration numbers of for compressed files with calls of this type:
```
iters = xmitgcm.mds_store._get_all_iternums(data_dir, file_prefixes=varname, file_format='*.??????????.data.shrunk')
```

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -749,7 +749,7 @@ def _concat_dicts(list_of_dicts):
     return result
 
 
-def _get_all_iternums(data_dir, file_prefixes=None, 
+def _get_all_iternums(data_dir, file_prefixes=None,
                       file_format='*.??????????.data'):
     """Scan a directory for all iteration number suffixes."""
     iternums = set()
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -749,13 +749,15 @@ def _concat_dicts(list_of_dicts):
     return result
 
 
-def _get_all_iternums(data_dir, file_prefixes=None):
+def _get_all_iternums(data_dir, file_prefixes=None, file_format='*.??????????.data'):
     """Scan a directory for all iteration number suffixes."""
     iternums = set()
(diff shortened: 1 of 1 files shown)


apatlpo:
Guys  what is the status on this?

rabernat:
So sorry for the delay. I had not realized there was an open PR here! Thanks for your reminder. 

Please add a test for this feature and update the documentation so we know what it actually does.

apatlpo:
This is a very minor modification of a hidden function.
Does it really deserve a test and specific documentation? (i'm fine with a plain 'yes' answer)
I can add more info in the docstring of the function also.

rabernat:
I see your point.

Could you at least explain in the description of the PR what this modification is for and why it is necessary for reading the compressed data?

xmitgcm/mds_store.py
```diff
@@ -749,13 +749,16 @@ def _concat_dicts(list_of_dicts):
     return result
 
 
-def _get_all_iternums(data_dir, file_prefixes=None):
+def _get_all_iternums(data_dir, file_prefixes=None,
+                      file_format='*.??????????.data'):
```
rabernat: How is the `file_format` keyword propagated from `open_mdsdataset` to this function?

apatlpo: It is not propagated from `open_mdsdataset`
The optional argument is intended for utility calls as described above.
Does it make sense?

rabernat:
Ok, now I understand.

In general, I did not expect that these private functions would be used in this way. But I am happy to merge this if it helps what you are doing.

FWIW, I would probably just manually specify the iternums in your project. There is no need to do a `glob` (which can be quite slow for thousands of files). You know the iternums exactly, since they have a specific predictable sequence.

apatlpo:
thanks for the merge and your advice

PR #81, opened 2018-06-08
Ignore iternums which only have pickup files

knaughten:
I found that xmitgcm breaks if you point it to a directory where some iternums only have pickup files, and no matching diagnostic files at the same timestep. Here is a small patch to ignore those iternums.

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -141,6 +141,7 @@ def open_mdsdataset(data_dir, grid_dir=None,
                     these_prefixes = _get_all_matching_prefixes(
                         data_dir, iternum, prefix
                     )
+                    # don't care about order
                     if set(these_prefixes) != set(first_prefixes):
                         raise IOError("Could not find the expected file "
                                       "prefixes %s at iternum %g. (Instead "
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -141,7 +141,6 @@ def open_mdsdataset(data_dir, grid_dir=None,
                     these_prefixes = _get_all_matching_prefixes(
                         data_dir, iternum, prefix
                     )
-                    # don't care about order
                     if set(these_prefixes) != set(first_prefixes):
                         raise IOError("Could not find the expected file "
                                       "prefixes %s at iternum %g. (Instead "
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks so much for this contribution!

I'm wondering if it would be possible to include a very short test for this in the test suite. Could you clarify how exactly "xmitgcm breaks" in this scenario? (i.e. Post the full error message and stack trace that you encounter.) Then we can figure out how to add a test for it.

knaughten:
Hey Ryan, here is an example: 

My workflow is that after each MITgcm run segment (as much as the model can do in a 24-hour queue limit) I convert all the diagnostic .data files to NetCDF using xmitgcm, then put the original files in a special folder so they don't get processed again after the next run segment. But I don't move any pickup files, as sometimes I need them for the restart.

Here's what happens if I point xmitgcm to a directory with some old permanent checkpoint files (eg pickup.0003682080.data) which don't correspond to any diagnostic files, which are starting at timestep 4075200. Here I'm using the commit from 6 April (commit f8ae6b3d7c1ecc58a955238bf76d600225ec0ad7), before I made any changes.

This is my code:

```
import sys
from xmitgcm import open_mdsdataset
ds=open_mdsdataset('.',delta_t=60,ref_date='1979-1-1 0:0:0')
```

and here is the error:

```
IOError                                   Traceback (most recent call last)
<ipython-input-3-df3d49ffabaf> in <module>()
----> 1 ds=open_mdsdataset('.',delta_t=60,ref_date='1979-1-1 0:0:0')

/work/n02/n02/kaight/python/xmitgcm/xmitgcm/mds_store.py in open_mdsdataset(data_dir, grid_dir, iters, prefix, read_grid, delta_t, ref_date, calendar, geometry, grid_vars_to_coords, swap_dims, endian, chunks, ignore_unknown_vars, default_dtype, nx, ny, nz, llc_method)
    148                                       "found %s)" % (repr(first_prefixes),
    149                                                      iternum,
--> 150                                                      repr(these_prefixes)))
    151 
    152                 # chunk at least by time

IOError: Could not find the expected file prefixes [] at iternum 4.0752e+06. (Instead found ['stateUVEL', 'stateEXF', 'state2D', 'stateWVEL', 'stateSALT', 'stateTHETA', 'stateICE', 'stateVVEL'])
```

Looks like what's happening in this case is that file prefixes are determined from the first iternum (which only has pickup files so they're ignored), and then the file prefixes from the diagnostic files don't match this (empty) list.

Issue #82, opened 2018-06-21
commit 0ca23c2 doesn't work on anaconda3


raphaeldussin:
doesn't pass the py.test:

E       NameError: name 'unicode' is not defined

removing or unicde on line 123 in mds_store.py works

rabernat:
Yes, this appears to be failing on travis as well!
https://travis-ci.org/xgcm/xmitgcm/jobs/389877039

The full stack trace is:
```
        # get frame info for history
        frame = inspect.currentframe()
        _, _, _, arg_values = inspect.getargvalues(frame)
        del arg_values['frame']
        function_name = inspect.getframeinfo(frame)[2]
    
        # auto-detect whether to swap dims
        if swap_dims is None:
            if read_grid == False:
                swap_dims = False
            else:
                swap_dims = False if geometry in ('llc', 'curvilinear') else True
    
        # some checks for argument consistency
        if swap_dims and not read_grid:
            raise ValueError("If swap_dims==True, read_grid must be True.")
    
        # if prefix is passed as a string, force it to be a list
>       if type(prefix) is str or unicode:
E       NameError: name 'unicode' is not defined
```

I don't know how this bug emerged.

Could you try to fix the error and submit a pull request? It seems like `unicode` needs to be imported.

raphaeldussin:
it seems that, since python3 is using unicode for all strings, the type unicode doesn't exist anymore.

https://stackoverflow.com/questions/6812031/how-to-make-unicode-string-with-python3

should we just remove the unicode section of the test?

rabernat:
We need to support python 2 and 3. 

The tests I showed you on my machine today were in python 3. 

This all worked a few weeks ago. I’m not sure what changed.

PR #83, opened 2018-06-22
fix the problem with testing type of prefix

raphaeldussin:
pass py.test in anaconda2 and anaconda3

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -28,13 +28,14 @@ from .utils import parse_meta_file, read_mds, parse_available_diagnostics
 
 # Python2/3 compatibility
 if (sys.version_info > (3, 0)):
-     stringtypes=[str]
+    stringtypes = [str]
 else:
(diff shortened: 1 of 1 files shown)


rabernat:
This looks perfect. Thanks!

I think this bug must have been introduced in #79. I wonder how we missed it then.

If you say "fixes #82" in your description or commit message, that issue will automatically be closed when this is merged. I will leave this open for a day to see if anyone else comments on it (unlikely) and then we can merge it.

Side note: probably not a good idea to develop new features on your local master branch. Use a feature branch instead. See [xmitgcm development workflow documentation](http://xmitgcm.readthedocs.io/en/latest/development.html#develpment-workflow).

raphaeldussin:
ok good to know for the commit message.
I did start a branch to tackle the rewriting of read_mds

edoddridge:
I’m likewise perplexed about how we didn’t catch this in #79. 
Thanks for fixing.

PR #84, opened 2018-06-22
extend capabilities of read_raw_data

raphaeldussin:
* possibility to read part of the file, with offset and partial_read
* choice of row/column major order

This will allow the refactor of read_mds by putting the np.fromfile and np.memmap
calls into one generic function

### DIFF
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -198,27 +198,26 @@ def test_parse_meta(tmpdir):
     for k, v in expected.items():
         assert result[k] == v
 
-
-def test_read_raw_data(tmpdir):
+@pytest.mark.parametrize("dtype", [np.dtype('f8'), np.dtype('f4'), np.dtype('i4')])
+def test_read_raw_data(tmpdir,dtype):
     """Check our utility for reading raw data."""
 
     from xmitgcm.utils import read_raw_data
     shape = (2, 4)
-    for dtype in [np.dtype('f8'), np.dtype('f4'), np.dtype('i4')]:
-        # create some test data
-        testdata = np.zeros(shape, dtype)
-        # write to a file
-        datafile = tmpdir.join("tmp.data")
-        datafile.write_binary(testdata.tobytes())
-        fname = str(datafile)
-        # now test the function
-        data = read_raw_data(fname, dtype, shape)
-        np.testing.assert_allclose(data, testdata)
-        # interestingly, memmaps are also ndarrays, but not vice versa
-        assert isinstance(data, np.ndarray) and not isinstance(data, np.memmap)
-        # check memmap
-        mdata = read_raw_data(fname, dtype, shape, use_mmap=True)
-        assert isinstance(mdata, np.memmap)
+    # create some test data
+    testdata = np.zeros(shape, dtype)
+    # write to a file
+    datafile = tmpdir.join("tmp.data")
+    datafile.write_binary(testdata.tobytes())
+    fname = str(datafile)
+    # now test the function
+    data = read_raw_data(fname, dtype, shape)
+    np.testing.assert_allclose(data, testdata)
+    # interestingly, memmaps are also ndarrays, but not vice versa
+    assert isinstance(data, np.ndarray) and not isinstance(data, np.memmap)
+    # check memmap
+    mdata = read_raw_data(fname, dtype, shape, use_mmap=True)
+    assert isinstance(mdata, np.memmap)
 
     # make sure errors are correct
     wrongshape = (2, 5)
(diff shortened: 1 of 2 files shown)


rabernat:
This looks like a nice simple extension of the read_raw_data function. I see how it will be useful in the future refactoring. Thanks! 

The new options need a test, however. The current test of `read_raw_data` is here:
https://github.com/xgcm/xmitgcm/blob/master/xmitgcm/test/test_mds_store.py#L202-L221

You can either extend that test or add a new test function that covers the new options.

raphaeldussin:
working on it. I have found some case where the error message is not very informative.
will catch those exception and set proper error message then resubmit PR with testing

raphaeldussin:
yeah finally!

xmitgcm/test/test_mds_store.py
```diff
@@ -225,8 +225,56 @@ def test_read_raw_data(tmpdir):
...
+            # offset is too big
+            _ = read_raw_data(fname, dtype, shape, offset=(
+                shape[0]*shape[1]*shape[2]*dtype.itemsize), partial_read=True)
+            _ = read_raw_data(fname, dtype, shape, offset=(
+                shape[0]*shape[1]*shape[2]*dtype.itemsize), partial_read=True,
+                use_mmap=True)
 
 # a meta test
+
+
```
rabernat: maybe remove these blank lines

xmitgcm/test/test_mds_store.py
```diff
@@ -225,8 +225,56 @@ def test_read_raw_data(tmpdir):
...
+            _ = read_raw_data(fname, dtype, shape_subset,
+                              offset=4, partial_read=False)
+            _ = read_raw_data(fname, dtype, shape_subset,
+                              offset=4, partial_read=False, use_mmap=True)
+            # offset is too big
+            _ = read_raw_data(fname, dtype, shape, offset=(
+                shape[0]*shape[1]*shape[2]*dtype.itemsize), partial_read=True)
+            _ = read_raw_data(fname, dtype, shape, offset=(
+                shape[0]*shape[1]*shape[2]*dtype.itemsize), partial_read=True,
+                use_mmap=True)
```
rabernat: This looks great! 👍

xmitgcm/utils.py
```diff
@@ -209,32 +210,47 @@ def read_raw_data(datafile, dtype, shape, use_mmap=False):
...
+                          (datafile,
+                           expected_number_of_bytes,
+                           actual_number_of_bytes))
     else:
-        # print("Reading %s using fromfile" % datafile)
-        d = np.fromfile(datafile, dtype)
-    d.shape = shape
-    return d
+        pass
+    assert(offset < actual_number_of_bytes), 'offset greater than filesize'
```
rabernat: should we raise an error here instead of assert?

assert should be used only for internal consistency checks. Is that what this is?

xmitgcm/test/test_mds_store.py
```diff
@@ -225,8 +225,56 @@ def test_read_raw_data(tmpdir):
     with pytest.raises(IOError):
         _ = read_raw_data(fname, dtype, wrongshape)
 
+    # test optional functionalities
+    shape = (5, 15, 10)
+    shape_subset = (15, 10)
+    for dtype in [np.dtype('f8'), np.dtype('f4'), np.dtype('i4')]:
```
rabernat: ok, one comment about pytest:
rather than doing a for loop within the test function, you could parameterize this dtype. Before the function, you could add
```python
    @pytest.mark.parametrize("dtype", [np.dtype('f8'), np.dtype('f4'), np.dtype('i4')])
    def test_read_raw_data(tmpdir, dtype):
```

xmitgcm/test/test_mds_store.py
raphaeldussin: fixed

xmitgcm/utils.py
rabernat: The way to resolve this is to put it back as an exception but add a test function that the error is raised.

Just do what you think is best and then merge.

rabernat:
I think this is ready to emrge, despite the codecov complaints.

Please do not make any new pull requests other than those related to the release. We need to make a release to mark the *current state* of xmitgcm, **before** making major changes. That is the whole point of having versions.

xmitgcm/utils.py
```diff
@@ -209,32 +210,53 @@ def read_raw_data(datafile, dtype, shape, use_mmap=False):
...
-        d = np.fromfile(datafile, dtype)
-    d.shape = shape
-    return d
+        pass
+
+    if offset < actual_number_of_bytes:
+        pass
+    else:
+        raise ValueError('bytes offset %g is greater than file size %g' %
+                         (offset, actual_number_of_bytes))
```
rabernat: Based on the codecov report, it looks like this is the error that is not getting tested:
https://codecov.io/gh/xgcm/xmitgcm/pull/84/src/xmitgcm/utils.py?before=xmitgcm/utils.py#L247

raphaeldussin: weird! because that's what I trying to test here : 

https://github.com/raphaeldussin/xmitgcm/blob/fe85c122d3f46c90b04ba054d7eba60cc6f08fc4/xmitgcm/test/test_mds_store.py#L266

rabernat: Ah! Now I understand why it doesn't work.

You need a separate `with` block for each exception you want to catch. The block exits as soon as it finds the exception. So only the first of the four `read_raw_data` calls in your `with pytest.raises(ValueError):` block is actually getting run.

Sorry I didn't catch that in my review.

raphaeldussin: thanks! Ok that make sense now!

rabernat:
Great!

PR #87, opened 2018-08-01
WIP refactoring of low level functions

raphaeldussin:
rewritting low level functions so they can handle cropped llc data and restart files
(mixed 3d/2d variables in files). All variables are outputted as dask arrays regardless
of geometry. Call order as follows : 

read_mds
  -> read_all_variables
     -> read_generic_data
       -> _read_xy_chunk
          -> read_raw_data

some changes are necessary in upper level functions read_mds and _MDSstore
and will be in next PR once this is merged.

WIP : need some tests. What do we want to test ? type = dask.array,...

### DIFF
--- a/xmitgcm/utils.py
+++ b/xmitgcm/utils.py
@@ -549,3 +549,218 @@ def _llc_data_shape(llc_id, nz=None):
     # should we accomodate multiple records?
     # no, not in this function
     return data_shape
+
+def read_all_variables(variable_list, metadata, use_mmap=True):
+    ''' return a dictionary of dask arrays
+    variable_list : list
+                    list of MITgcm variables, from fldList in .meta
+    metadata      : dict
+                    internal metadata for binary file
+    use_mmap : bool, optional
+               Whether to read the data using a numpy.memmap
+    '''
+    out = []
+    for variable in variable_list:
+        out.append(read_generic_data(variable, metadata, use_mmap=use_mmap))
+
+    return out
+
+def read_generic_data(variable, metadata, use_mmap=True):
+    ''' return dask array for variable using the given metadata
+    variable : string
+               name of the variable to read
+    metadata : dict
+               internal metadata for binary file
+    use_mmap : bool, optional
+               Whether to read the data using a numpy.memmap
+    '''
+
+    if (metadata['nx'] == 1) and (metadata['ny'] == 1) and
+       (len(metadata['vars']) == 1):
+        # vertical coordinate
+        data_raw = read_raw_data(metadata['filename'], metadata['dtype'],
+                                 (metadata['nz'],), use_mmap=use_mmap, offset=0,
+                                 order='C', partial_read=False)
+
+        shape = (metadata['nt'], metadata['nz'], 1,
+                 metadata['nx'], metadata['nx'])
+        data_raw = np.reshape(data_raw,shape)
+        chunks = (metadata['nt'], 1, 1,
+                  metadata['nx'], metadata['nx'])
+        data = dsa.from_array(data_raw, chunks=chunks)
+
+    else:
+        if metadata['has_faces']:
+            def load_chunk(face, lev, rec):
+                return _read_xy_chunk(variable,metadata,rec=rec,lev=lev,
+                                      face=face,use_mmap=use_mmap)[None, None, None]
+
+            chunks = (1, 1, 1, metadata['nx'], metadata['nx'])
+            shape = (metadata['nt'], metadata['nz'], len(metadata['face_facets']),
+                     metadata['nx'], metadata['nx'])
+            name = 'llc-' + tokenize(metadata['filename'])
+
+            dsk = {(name, rec, lev, face, 0, 0): (load_chunk, face,
+                                                     lev , rec)
+                     for face in range(len(metadata['face_facets']))
+                     for lev in range(metadata['nz'])
+                     for rec in range(metadata['nt'])}
+
+            data = dsa.Array(dsk, name, chunks, dtype=metadata['dtype'], shape=shape)
+        else:
+            def load_chunk(lev, rec):
+                return _read_xy_chunk(variable,metadata,rec=rec,lev=lev,
+                                      face=0,use_mmap=use_mmap)[None, None]
+
+            chunks = (1, 1, metadata['ny'], metadata['nx'])
+            shape = (metadata['nt'], metadata['nz'],
+                     metadata['ny'], metadata['nx'])
+            name = 'reg-' + tokenize(metadata['filename'])
+
+            dsk = {(name, rec, lev, 0, 0): (load_chunk, lev , rec)
+                     for lev in range(metadata['nz'])
+                     for rec in range(metadata['nt'])}
+
+            data = dsa.Array(dsk, name, chunks, dtype=metadata['dtype'], shape=shape)
+
+    return data
+
+def _read_xy_chunk(variable, metadata, rec=0, lev=0, face=0, use_mmap=False):
+    ''' read a 2d chunk along (x,y)
+    variable : string
+               name of the variable to read
+    metadata : dict
+               metadata for binary file
+    rec      : integer, optional
+               time record to read (default=0)
+    lev      : integer, optional
+               vertical level to read (default=0)
+    face     : integer, optional
+               face to read for llc configurations (default=0)
+    use_mmap : bool, optional
+               Whether to read the data using a numpy.memmap
+    '''
+
+    # size of the data element
+    nbytes = metadata['dtype'].itemsize
+    # byte order
+    metadata['dtype'] = metadata['dtype'].newbyteorder(metadata['endian'])
+    # find index of variable
+    idx_var = metadata['vars'].index(variable)
+
+    # 1. compute offset_variable, init to zero
+    offset_vars = 0
+    # loop on variables before the one to read
+    for jvar in np.arange(idx_var):
+        # inspect its dimensions
+        dims = metadata['dims_vars'][jvar]
+        # compute the byte size of this variable
+        nbytes_thisvar=1*nbytes
+        for dim in dims:
+            nbytes_thisvar=nbytes_thisvar*metadata[dim]
+        # update offset from previous variables
+        offset_vars=offset_vars+nbytes_thisvar
+
+    # 2. get dimensions of desired variable
+    dims = metadata['dims_vars'][idx_var]
+    # set values to 1 by default
+    nx=1 ; ny=1 ; nz=1 ; nt=1
+    # update with true values
+    if 'nx' in dims : nx=metadata['nx']
+    if 'ny' in dims : ny=metadata['ny']
+    if 'nz' in dims : nz=metadata['nz']
+    if 'nt' in dims : nt=metadata['nt']
+
+    # 3. compute offset from previous records of current variable
+    if ( rec > nt-1):
+        raise ValueError("time record %g greater than number of records %g" %
+                         (rec, nt))
+    else:
+        offset_timerecords = rec * nz * ny * nx * nbytes
+
+    # 4. compute offset from previous vertical levels of current variable
+    if ( lev > nz-1):
+        raise ValueError("level %g is greater than number of levels %g" %
+                         (lev, nz))
+    else:
+        offset_verticallevels = lev * ny * nx * nbytes
+
+    # 5. compute the offset of the previous variables, records and levels
+    offset = offset_vars + offset_timerecords + offset_verticallevels
+
+    # 6. compute offset due to faces
+    if metadata['has_faces']:
+        # determin which facet the face belong to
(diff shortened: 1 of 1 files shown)


rabernat:
FYI, you can avoid these stickler comments by running [flake8](http://flake8.pycqa.org/en/latest/) on your code beforehand.

xmitgcm/utils.py
```diff
@@ -549,3 +549,236 @@ def _llc_data_shape(llc_id, nz=None):
...
+    if 'pad_before_y' in metadata:
+        if metadata['has_faces']:
+            nypad_before = metadata['pad_before_y'][facet_origin]
+        else:
+            nypad_before = metadata['pad_before_y']
+
+        pad_before = dsa.from_array(
+            np.zeros((nypad_before, nx)), chunks=(nypad_before, nx))
+        ddata_padded_before = dsa.concatenate(
+            (pad_before, ddata_interior), axis=0)
```
rabernat: These calls to zeros / concatenate should be refactored to use some sort of duck-operations, selecting either dask or numpy functions depending on the user's desired format. See
https://github.com/xgcm/xgcm/blob/master/xgcm/duck_array_ops.py

xmitgcm/utils.py
```diff
@@ -549,3 +549,236 @@ def _llc_data_shape(llc_id, nz=None):
...
+            nypad_after = metadata['pad_after_y'][facet_origin]
+        else:
+            nypad_after = metadata['pad_after_y']
+
+        pad_after = dsa.from_array(
+            np.zeros((nypad_after, nx)), chunks=(nypad_after, nx))
+        ddata_padded_after = dsa.concatenate(
+            (ddata_padded_before, pad_after), axis=0)
+    else:
+        ddata_padded_after = ddata_padded_before
```
rabernat: I would recommend separating out the padding stuff into a standalone function which you then call from here.

rabernat:
Numpy memmap docs:
https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.memmap.html

xmitgcm/utils.py
```diff
@@ -549,3 +549,236 @@ def _llc_data_shape(llc_id, nz=None):
...
+                     metadata['nx'], metadata['nx'])
+            name = 'llc-' + tokenize(metadata['filename'])
+
+            dsk = {(name, rec, lev, face, 0, 0): (load_chunk, face,
+                                                  lev, rec)
+                   for face in range(len(metadata['face_facets']))
+                   for lev in range(metadata['nz'])
+                   for rec in range(metadata['nt'])}
+
+            data = dsa.Array(dsk, name, chunks,
```
rabernat: So this construction of dask arrays is already lazy. What happens when you try to actually access the data is that `load_chunk` gets called and the result is converted to a numpy array. 

This means that `load chunk` should just return numpy arrays.

xmitgcm/test/test_mds_store.py
```diff
@@ -386,31 +386,323 @@ def test_read_raw_data_llc(llc_mds_datadirs, method, memmap):
...
+                              'dims_vars': dimsvar})
+
+        for kface in np.arange(13):
+            data = _read_xy_chunk('MXLDEPTH', file_metadata, face=kface,
+                                  use_mmap=memmap)
+
+        if memmap:
+            assert type(data) == np.memmap
+        else:
+            assert type(data) == np.ndarray
```
rabernat: For type assertions, probably best to say
```
assert isinstance(data, np.ndarray)
```
see https://lat.sk/2017/03/type-checking-assertions-exceptions-mypy/ for some good advice

xmitgcm/test/test_mds_store.py
```diff
@@ -386,31 +386,323 @@ def test_read_raw_data_llc(llc_mds_datadirs, method, memmap):
...
+    assert data_padded[0, 0] == 1
+    assert data_padded[1, 1] == 4
+
+    data_padded = _pad_array(data, file_metadata, face=1)
+    assert type(data_padded) == np.ndarray
+    assert data_padded.shape == (5, 2)
+    assert data_padded.min() == 0
+    assert data_padded.max() == 4
+    assert data_padded[0, 0] == 1
+    assert data_padded[1, 1] == 4
```
rabernat: Very comprehensive tests of padding function! Nice!

Does the padding function operate always on numpy arrays? Or can it take dask arrays?

xmitgcm/utils.py
```diff
@@ -549,3 +549,285 @@ def _llc_data_shape(llc_id, nz=None):
...
+            shape = (file_metadata['nt'], file_metadata['nz'],
+                     file_metadata['ny'], file_metadata['nx'])
+            name = 'reg-' + tokenize(file_metadata['filename'])
+
+            dsk = {(name, rec, lev, 0, 0): (load_chunk, lev, rec)
+                   for lev in range(file_metadata['nz'])
+                   for rec in range(file_metadata['nt'])}
+
+            data = dsa.Array(dsk, name, chunks,
+                             dtype=file_metadata['dtype'], shape=shape)
```
rabernat: Since this last line is common to both above code blocks, perhaps it should be placed outside the if / else statement.

xmitgcm/utils.py
```diff
@@ -549,3 +549,285 @@ def _llc_data_shape(llc_id, nz=None):
...
+    data          : numpy array or memmap
+                    input data
+    file_metadata : dict
+                    metadata for file
+    face          : int, optional
+                    llc face if applicable
+
+    Returns
+    -------
+    numpy.array
```
rabernat: Ok, I see...it is clear that this function only operates on numpy arrays / memmaps (not dask arrays).

xmitgcm/utils.py
```diff
@@ -549,3 +549,285 @@ def _llc_data_shape(llc_id, nz=None):
...
+    is necessary, the function preserves its type. Otherwise, the concatenate
+    forces it to load into memory.
+
+    Parameters
+    ----------
+
+    data          : numpy array or memmap
+                    input data
+    file_metadata : dict
+                    metadata for file
```
rabernat: Again, I am concerned that it is not clear what is supposed to be in file_metadata. This function would be easier to understand if these options were explicit keywords, rather than hidden in a dictionary.

xmitgcm/test/test_mds_store.py
raphaeldussin: corrected

xmitgcm/test/test_mds_store.py
raphaeldussin: 👍

xmitgcm/utils.py
raphaeldussin: good catch!

xmitgcm/utils.py
raphaeldussin: or return a memmap if not padding is applied, corrected

xmitgcm/utils.py
raphaeldussin: file_metadata is gonna be the union of the metadata in *.meta and an extra_metadata passed
as an arg of read_mds. Will describe in read_mds the expected content of extra_metadata

xmitgcm/utils.py
```diff
@@ -549,3 +549,283 @@ def _llc_data_shape(llc_id, nz=None):
...
+
+    out = []
+    for variable in variable_list:
+        out.append(read_generic_data(variable, file_metadata,
+                                     use_mmap=use_mmap))
+
+    return out
+
+
+def read_generic_data(variable, file_metadata, use_mmap=False):
```
rabernat: use_dask=True

xmitgcm/utils.py
```diff
@@ -549,3 +549,283 @@ def _llc_data_shape(llc_id, nz=None):
...
+            shape = (file_metadata['nt'], file_metadata['nz'],
+                     file_metadata['ny'], file_metadata['nx'])
+            name = 'reg-' + tokenize(file_metadata['filename'])
+
+            dsk = {(name, rec, lev, 0, 0): (load_chunk, lev, rec)
+                   for lev in range(file_metadata['nz'])
+                   for rec in range(file_metadata['nt'])}
+
+        data = dsa.Array(dsk, name, chunks,
+                         dtype=file_metadata['dtype'], shape=shape)
```
rabernat:
```python
if not use_dask:
    data = data.load()
```

rabernat: you could even use the serial scheduler if you want to avoid triggering parallel computations

xmitgcm/utils.py
```diff
@@ -549,3 +549,283 @@ def _llc_data_shape(llc_id, nz=None):
...
+                                                  lev, rec)
+                   for face in range(len(file_metadata['face_facets']))
+                   for lev in range(file_metadata['nz'])
+                   for rec in range(file_metadata['nt'])}
+
+        else:
+            def load_chunk(lev, rec):
+                return _read_xy_chunk(variable, file_metadata,
+                                      rec=rec, lev=lev,
+                                      face=0, use_mmap=use_mmap)[None, None]
```
rabernat: Here it would be great if we could have more than one chunk in the vertical...up to `nz`.

rabernat: this option would only work for non faces grids

raphaeldussin: doing a functions that reads an arbitraty number of vertical levels for the chunk is not trivial. Can we do one that reads the whole 3d field so we don't loose capabilities and save the arbitrary number of levels for later ?

rabernat:
Just checking in on this. It seems like there are some test failures happening related to memmap.

xmitgcm/utils.py
```diff
@@ -549,3 +549,424 @@ def _llc_data_shape(llc_id, nz=None):
...
+    data = dsa.Array(dsk, name, chunks,
+                     dtype=file_metadata['dtype'], shape=shape)
+
+    if not use_dask:
+        data = data.compute()
+
+    return data
+
+
+def _read_3d_chunk(variable, file_metadata, rec=0, use_mmap=False):
```
rabernat: Maybe `_read_xyz_chunk` would be more consistent with the other function name?

xmitgcm/utils.py
```diff
@@ -549,3 +549,424 @@ def _llc_data_shape(llc_id, nz=None):
...
+    # in conventional grids, it's in C
+    order = 'C'
+
+    # 5. Do the actual read
+    data = read_raw_data(file_metadata['filename'],
+                         file_metadata['datatype'],
+                         shape, use_mmap=use_mmap, offset=offset,
+                         order=order, partial_read=partial_read)
+
+    return data
```
rabernat: I think some refactoring is needed here. The above function is so similar to `_read_xy_chunk`. Surely there is a way to consolidate much of the logic rather than copy / paste the code from one function to another.

raphaeldussin:
the memmap errors are linked to an issue I mentioned earlier. For some reason, read_small_chunks and read_big_chunks don't behave the same way when dask=False. read_big_chunk returns the correct type (memmap on ndarray) as asked in the args. However read_small_chunks always returns a ndarray when dask=False. I don't know if that's related to how the dask graph is built. Any ideas are welcome, I'm stuck on this issue.

I weighted the options between duplicating read_xy_chunk into read_xyz_chunk or making it one big
function. As much as I hate duplicating code, adding the big/small chunk option to read_xy_chunk
(to make it a read_chunk basically) is gonna make the function longer and really painful to read.

raphaeldussin:
All test are passing if we lift the test on memmap type : 

diff --git a/xmitgcm/test/test_mds_store.py b/xmitgcm/test/test_mds_store.py
index a5fbb6d..fd4da85 100644
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -565,7 +565,8 @@ def test_read_small_chunks(all_mds_datadirs, memmap, usedask):
         data.compute()  # check accessing data works
     else:
         if memmap:
-            assert isinstance(data, np.memmap)
+            #assert isinstance(data, np.memmap)
+            assert isinstance(data, np.ndarray)
         else:
             assert isinstance(data, np.ndarray)

@@ -735,7 +736,7 @@ def test_read_all_variables(all_mds_datadirs, memmap, usedask):  # , chunks):
         assert isinstance(dataset[0], dask.array.core.Array)
     else:
         if memmap:
-            assert isinstance(dataset[0], np.memmap)
+            assert isinstance(dataset[0], np.ndarray)
         else:
             assert isinstance(dataset[0], np.ndarray)

@@ -763,7 +764,7 @@ def test_read_all_variables(all_mds_datadirs, memmap, usedask):  # , chunks):
         assert isinstance(dataset[0], dask.array.core.Array)
     else:
         if memmap:
-            assert isinstance(dataset[0], np.memmap)
+            assert isinstance(dataset[0], np.ndarray)
         else:
             assert isinstance(dataset[0], np.ndarray)

ASTE is gonna break that test anyway because of the padding of arrays that will trigger the
evaluation of the memmap.

PR #93, opened 2018-08-31
clean refactored read_mds

raphaeldussin:


xmitgcm/mds_store.py
```diff
@@ -595,7 +595,8 @@ def load_from_prefix(self, prefix, iternum=None):
                 # The answer appears to be yes: 2D (x|y,z) data retains the
                 # missing dimension as an axis of length 1.
                 dims = dims[1:]
-            elif len(dims) == 1 and (data.ndim == 2 or data.ndim == 3):
+            elif len(dims) == 1 and (data.ndim == 2 or data.ndim == 3 or
+                                     data.ndim == 4):
```
rabernat: `data.ndim > 1` ?

xmitgcm/test/test_mds_store.py
```diff
@@ -296,24 +296,63 @@ def test_read_mds(all_mds_datadirs):
...
+        emeta = {'has_faces': False}
+    res = read_mds(basename, chunks="small", dask_delayed=False,
+                   use_mmap=False, extra_metadata=emeta)
+    assert isinstance(res, dict)
+    assert prefix in res
+    assert isinstance(res[prefix], np.ndarray)
 
     # make sure endianness works
-    testval = res.newbyteorder('<')[0,0]
-    res_endian = read_mds(basename, force_dict=False, use_mmap=False,
```
rabernat: It now looks like we are not testing the `force_dict` option anywhere.

xmitgcm/utils.py
```diff
@@ -105,17 +104,66 @@ def read_mds(fname, iternum=None, use_mmap=True, force_dict=True, endian='>',
...
+                                                 after array
+        * facet_order : row/column major order of this facet
+
+        -list of len=nfaces:
+
+        * face_facets : facet of origin for this face
+
+        * face_offsets : position of the face in the facet (0 = start)
+
+        * transpose_face : transpose the data for this face
```
rabernat: Have you checked that this documentation string formats correctly when you build the docs?

xmitgcm/utils.py
```diff
@@ -136,64 +184,112 @@ def read_mds(fname, iternum=None, use_mmap=True, force_dict=True, endian='>',
...
+        llc = {'has_faces': True, 'ny': 13*nx, 'nx': nx,
+               'ny_facets': [3*nx, 3*nx, nx, 3*nx, 3*nx],
+               'face_facets': [0, 0, 0, 1, 1, 1, 2, 3, 3, 3, 4, 4, 4],
+               'facet_orders': ['C', 'C', 'C', 'F', 'F'],
+               'face_offsets': [0, 1, 2, 0, 1, 2, 0, 0, 1, 2, 0, 1, 2],
+               'transpose_face': [False, False, False,
+                                  False, False, False, False,
+                                  True, True, True, True, True, True]}
+        file_metadata.update(llc)
+    # --------------- /LEGACY --------------------------
```
rabernat: just a thought...maybe we need library of `extra_metadata` for different configurations (llc, aste, cs, etc.)

xmitgcm/utils.py
```diff
@@ -136,64 +184,112 @@ def read_mds(fname, iternum=None, use_mmap=True, force_dict=True, endian='>',
...
+
+    # it is possible to override the values of nx, ny, nz from extra_metadata
+    # (needed for bug meta file ASTE) except if those are = 1 (vertical coord)
+    # where we override by values found in meta file
+    for dim in ['nx', 'ny', 'nz']:
+        if metadata[dim] == 1:
+            file_metadata.update({dim: 1})
+
+    use_dask = True if dask_delayed else False
```
rabernat: should we deprecate one in favor of the other so we used consistent keywords throughout xmitgcm?

xmitgcm/test/test_mds_store.py
raphaeldussin: I have deprecated this option since read_mds now always returns a dict. Is there cases when we need to return a different type ?

xmitgcm/utils.py
raphaeldussin: I like the idea! what about a function that returns the extra_metadata dictionary given a short list of arguments, e.g. : 

extra_llc = xmitgcm.obtain_extra_metadata(domain='llc', nx=90)
extra_aste = xmitgcm.obtain_extra_metadata(domain='aste', nx=270)

xmitgcm/utils.py
raphaeldussin: Absolutely! I started working with the use_dask bool not to interfere with dask_delayed.
The quickest way to do would be to replace dask_delayed by use_dask in read_mds, but I am fine with replacing the use_dask flags by dask_delayed if it's easier to understand. thoughts?

xmitgcm/utils.py
raphaeldussin: fixed

raphaeldussin:
@rabernat stickler is nitpicking on line lengths that are barely longer than 80 characters and not
trivial to change. Everything else works fine and all comments have been addressed.

xmitgcm/test/test_utils.py
```diff
@@ -0,0 +1,1003 @@
...
+    tar.extractall(target_dir)
+    tar.close()
+    # subdirectory where file should have been untarred.
+    # assumes the directory is the same name as the tar file itself.
+    # e.g. testdata.tar.gz --> testdata/
+    fulldir = os.path.join(target_dir, basename)
+    if not os.path.exists(fulldir):
+        raise IOError('Could not find tar file output dir %s' % fulldir)
+    # the actual data lives in a file called testdata
+    return fulldir
```
rabernat: All of the above code is cut and pasted verbatim from test_mds_store, so now it appears identically in two files. You know that this is not okay. We need to move all the common metadata, fixtures, utility functions, etc. that are used by the different test modules into its own module and then import it. Otherwise the code becomes impossible to maintain.

xmitgcm/utils.py
```diff
@@ -95,27 +96,77 @@ def read_mds(fname, iternum=None, use_mmap=True, force_dict=True, endian='>',
...
+
+        * list of len=nfaces:
+
+        #. face_facets : facet of origin for this face
+
+        #. face_offsets : position of the face in the facet (0 = start)
+
+        #. transpose_face : transpose the data for this face
+
+    chunks : {'big', 'small'}
```
rabernat: I really want to name these '3D' and '2D', rather than 'big' and 'small'.

rabernat:
Should we turn off this stickler? Or modify its behavior? It sure is annoying.

rabernat:
Ok I think we are good here.

Issue #95, opened 2018-09-05
Errors identifying files in run directory


rabernat:
On the MITgcm mailing list, @mjlosch reported some problems parsing the files in his run directory

>the diagnostics package produces 
a file cheapaml at iteration 11 and iceDiag at 12. I acknowledge that 
this is hard to catch in a generic reading software. But when I remove 
the cheapaml (and iceDiag) files and retry I still get:
>
> OSError: Could not find the expected file prefixes ['W', 'Eta', 
'ice_Tice1', 'ice_iceH', 'ice_Qice2', 'ice_snowH', 'S', 'ice_Tsrf', 
'ice_Qice1', 'T', 'ice_snowAge', 'ice_fract', 'V', 'VICE', 'U', 
'ice_Tice2', 'UICE'] at iternum 12. (Instead found ['ice_Tsrf', 
'ice_Qice1', 'T', 'V', 'ice_Tice2', 'Eta', 'Qnet', 'ice_frwAtm', 'Qsw', 
'EmPmR', 'ice_flxAtm', 'W', 'PH', 'ice_Tice1', 'ice_Qice2', 'FU', 
'ice_snowH', 'VICE', 'U', 'PHL', 'ice_iceH', 'S', 'FV', 'ice_snowAge', 
'ice_fract', 'UICE'])
>
> To me this looks like the same list of files but in different order. 
ignore_unknown_vars=True does not work.

He then elaborated

```
cd MITgcm/verification
./testreport -t global_ocean.cs32x15
cd global_ocean.cs32x15/tr_run.icedyn
ipython —pylab
import xmitgcm as xm
d = xm.open_mdsdataset('./‘,geometry='curvilinear')
```
gives
```
OSError: Could not find the expected file prefixes ['ice_Tice2', 'W', 'ice_fract', 'U', 'ice_snowH', 
'ice_Qice2', 'ice_Tsrf', 'ice_Qice1', 'Eta', 'ice_Tice1', 'V', 'ice_iceH', 'S', 'ice_snowAge', 'T'] at iternum 72010. (Instead found ['ice_iceH', 'ice_Tice2', 'U', 'ice_fract', 'W', 'ice_flxAtm', 'PHL', 'ice_Qice2', 'ice_frwAtm', 'V', 'ice_Tsrf', 'Eta', 'ice_Tice1', 'ice_snowH', 'ice_Qice1', 'S', 'ice_snowAge', 'T', 'PH'])
```

```
import glob
[...]

PR #96, opened 2018-09-06
Add seaice and thsice native output

mjlosch:
This is a first attempt to add the meta-data of the native output variables of pkg/seaice and pkg/thsice. It's like incomplete. For example, there is `thSIce_tave` which contains all tave-variables (see [this code](https://github.com/MITgcm/MITgcm/blob/2df144911f561b69553f329ff114d79f48fbdf32/pkg/thsice/thsice_output.F#L146)) and I don't know how to add that in a consistent way. This is supposed to fix most of issue #95


mjlosch:
Do I see this right? "stickler-ci" changed the indenting of my new code (send me 17 emails) and then complains about lines being too long? Are my indent convention (I use the default of emacs, like 3 spaces) not OK?

Issue #99, opened 2018-09-19
dask token problems


rabernat:
It is important that each variable created in xmitgcm have a unique dask token. @cspencerjones and I discovered this is not the case for variables coming from the same file with the latest version of xmitgcm. Consider this example from the global_oce_latlon test dataset.

```python
import xmitgcm
import xarray as xr

ddir = '/home/rpa/xmitgcm/test_data/global_oce_latlon'
prefix = ['DiagSurf-TS']
iters = [39600]
ds = xmitgcm.open_mdsdataset(ddir, prefix=prefix, iters=iters)
```
The data vars are
```
Data variables:
    TFLUX                 (time, YC, XC) >f4 dask.array<shape=(1, 40, 90), chunksize=(1, 40, 90)>
    surForcS              (time, YC, XC) >f4 dask.array<shape=(1, 40, 90), chunksize=(1, 40, 90)>
    surForcT              (time, YC, XC) >f4 dask.array<shape=(1, 40, 90), chunksize=(1, 40, 90)>
    SFLUX                 (time, YC, XC) >f4 dask.array<shape=(1, 40, 90), chunksize=(1, 40, 90)>
```

All the variables come from the same file (`DiagSurf-TS`). They work ok if we operate on them individually
```python
>>> print(ds.TFLUX.mean().load())
>>> print(ds.SFLUX.mean().load())
<xarray.DataArray 'TFLUX' ()>
array(-0.793143630027771, dtype=float32)
<xarray.DataArray 'SFLUX' ()>
array(-4.2861000110860914e-05, dtype=float32)
```

However, if we examine the keys from the dask graph, we notice something troubling:
```python
>>> list(ds.SFLUX.data.dask.keys())
[('getitem-81147f30dcd33bc7374d1ccba76a3840', 0, 0, 0),
 ('reg-88fb9d5b26c498736d854640a64c6128', 0, 0, 0, 0),
 ('getitem-3e2e9518ba0995fd71b85387c30631c3', 0, 0, 0),
 ('getitem-9bf4dd6e020d038f104d18f17c47139a', 0, 0)]
>>> list(ds.TFLUX.data.dask.keys())
[('getitem-81147f30dcd33bc7374d1ccba76a3840', 0, 0, 0),
 ('reg-88fb9d5b26c498736d854640a64c6128', 0, 0, 0, 0),
 ('getitem-3e2e9518ba0995fd71b85387c30631c3', 0, 0, 0),
[...]

PR #101, opened 2018-09-19
fix token bug

rabernat:
fixes: #99


raphaeldussin:
I had a PR for a fix with #100 
I'll close it then

rabernat:
Ah they crossed in the ether! I pushed this just before leaving work. Great minds...😉

The test here is not quite right yet though

rabernat:
LGTM. @raphaeldussin?

PR #105, opened 2018-10-05
support tracers CFC/BLING

raphaeldussin:
for some reason, some packages use TRAC instead of PTRACER in the diagnostics files.
This allows to read results from CFC and BLING


raphaeldussin:
turns out having available_diagnostics.log in the diags directory solves the problem. closing

Issue #108, opened 2018-10-11
Read in "special" grid data


timothyas:
There are some MITgcm files which get printed out that are "grid-like" but only show up in special cases, for instance:

- rLowC (etc) gets printed from write_grid when `debugLevel` >= `debugLevC`
- maskInC/W/S get printed when pkg/obcs is being used 
- AHybSigF when pressure coordinates are used (although I'm totally unfamiliar with this...)
- maskCtrlC/W/S get printed when pkg/ctrl is being used 
- probably others ... 

These are however separate from typical data variables because they don't have the ten digit time stamp, so adding them is not as simple as adding to `package_state_variables` which are read in as data vars. I would propose to do this by modifying `_get_all_grid_variables` so that after first grabbing the standard grid variables, it looks for these optional files with an adapted version of `_get_all_matching_prefixes`. 

Additionally, this seems related to #44, and could be addressed here as well.

I'm happy to implement this as I'm finally breaking my matlab addiction, but I wanted to see what you think before I code up a PR since I am a bit of a python and xmitgcm newbie.

rabernat:
Sorry for the slow reply here.

Yes, we definitely would want the ability to read all this stuff! PR welcome!

raphaeldussin:
fixed by #110 ?

PR #110, opened 2018-11-10
Extra grid vars

timothyas:
This PR specifically address #108 as follows: 

- "extra" grid variables printed from MITgcm/model/src/write_grid.F added to variables.py as one `OrderedDict` called `extra_grid_variables`
- sigma coordinate coefficients are skipped because it's not clear to me where each of these variables exist, but happy to add them if any of you guys have experience
- added routines to mds_store.py 
   - _get_extra_grid_variables : this grabs the list of available variables
   - _get_extra_grid_prefixes : looks for prefixes in data_dir and returns a list if any of the extra_grid_vars are found

- I assumed these variables are found in data_dir rather than grid_dir because these files are output at run time, and are useful for debugging, while I would picture a users grid_dir to be somewhere on their machine that doesn't change (somewhere other than the run directory). This could be easily fixed if you guys disagree.

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -508,8 +508,7 @@ class _MDSDataStore(xr.backends.common.AbstractDataStore):
         if read_grid:
             prefixes = prefixes + list(self._all_grid_variables.keys())
             prefixes = (prefixes + 
-                    _get_extra_grid_prefixes(data_dir, 
-                                                 file_prefixes))
+                    _get_extra_grid_prefixes(data_dir))
 
         # add data files
         prefixes = (prefixes +
@@ -864,7 +863,7 @@ def _is_pickup_prefix(prefix):
             return True
     return False
 
-def _get_extra_grid_prefixes(data_dir, file_prefixes=None):
+def _get_extra_grid_prefixes(data_dir):
     """Scan a directory and return all file prefixes for extra grid files."""
     prefixes = set()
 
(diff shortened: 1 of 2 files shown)


rabernat:
@raphaeldussin - would be great if you could review this.

timothyas:
Hey @raphaeldussin, I could do that too. I was keeping these separate because xmitgcm adds _all_grid_variables to the list of prefixes to look for, and the extra_grid_variables which I added are not always printed out, so most users would run into an error. 

However, I could make it so that _get_all_grid_variables only adds these extra variables to the list if the prefixes exist within grid_dir (in this case, _get_all_grid_variables will need an extra argument for the directory in which these files exist).  

I did not originally do this because it seemed to conflict the definitions of prefix and variables (because it looks for prefixes within a function that defines variables to the user's geometry), which seem to be purposefully separate throughout the package. But this wouldn't be difficult to change if that doesn't matter.

raphaeldussin:
@timothyas LGTM! @rabernat are you ok to merge?

xmitgcm/mds_store.py
```diff
@@ -741,6 +748,21 @@ def _get_all_grid_variables(geometry, layers={}):
     return metadata
 
 
+def _get_extra_grid_variables(grid_dir):
+    """Scan a directory and return all file prefixes for extra grid files.
+       Then return the variable information for each of these"""
+    extra_grid = {}
+
+    all_datafiles = glob(os.path.join(grid_dir, '*.data'))
```
rabernat: In some cases, these glob operations can be very slow. For the future, we should refactor the code so that grid_dir and data_dir are globbed exactly once and the results cached somewhere.

timothyas:
No problem! Thanks guys!

timothyas:
Closes #110

PR #115, opened 2019-01-10
transfer test data to figshare

raphaeldussin:
* removed all archives for test and moved them to figshare:

https://figshare.com/collections/xmitgcm_test_data/4362224

* add download function


raphaeldussin:
@rabernat good to go ?

rabernat:
This looks like a great start! I really like your design.

I would encourage you to look at the way xarray handles tutorial datasets:
https://github.com/pydata/xarray/blob/master/xarray/tutorial.py#L28

There are two ideas we can borrow from this, which I would like to see in xmitgcm
- [ ] use a specific cache directory (e.g. `$HOME/.xmitgcm-test-data/`) rather than the current working directory
- [ ] provide md5 checksums for the archive files to verify integrity (I guess the checksums would have to be added to the figshare archives)

raphaeldussin:
@rabernat  figshare provides a md5 for the files stored there so we just need to add that checksum in the experiment info.

I like the cache idea, I am just not sure $HOME is the right default location. Many HPC systems have very small disk quotas and some of our test cases might become big as resolution increases,...
figshare allows up to 5GB per file, which is comparable to the disk quota of some systems.
What about leaving the xmitgcm/test to be the default and add an option to add the cache directory
through an environment variable ?

rabernat:
👍 to $HOME by default + environment variable to customize. Make sure to document this in the docs.

rabernat:
Looks great!

FYI, the large .tar.gz files will continue to live in the repository because they are part of the commit history. If we want to reduce the size of the repository, we will have to edit the commit history using a technique like this: https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-git-repository

PR #118, opened 2019-01-17
Write binary files from xmitgcm

raphaeldussin:
allow to rebuild facets, rebuild compact form and write to file

testing almost finished

### DIFF
--- a/xmitgcm/utils.py
+++ b/xmitgcm/utils.py
@@ -1335,50 +1335,109 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
     return grid
 
 
-#def write_to_binary(da, fileout, precision='single', extra_metadata=None):
-#    ''' write variable from dataset ds to fileout with precision '''
+def find_concat_dim_facet(da, facet, extra_metadata):
+    """ find along which dimension to concatenate facet """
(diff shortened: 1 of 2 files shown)


xmitgcm/utils.py
```diff
@@ -1332,3 +1333,246 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
...
+            tmp = flatdata.astype(np.dtype('f')).byteswap(True).tobytes()
+        # big endian systems are not common place
+        else:  # pragma: no cover
+            tmp = flatdata.astype(np.dtype('f')).tobytes()
+    elif precision == 'double':
+        if sys.byteorder == 'little':
+            tmp = flatdata.astype(np.dtype('d')).byteswap(True).tobytes()
+        # big endian systems are not common place
+        else:  # pragma: no cover
+            tmp = flatdata.astype(np.dtype('d')).tobytes()
```
rabernat: Why introduce a new keyword called `precision`? Elsewhere in xmitgcm we just specify numpy dtypes explicitly?

Then you could just say 
```python
tmp = flatdata.astype(dtype).tobytes()
```

and forget about the 10 lines above

PR #119, opened 2019-02-18
more generic 3d function

raphaeldussin:


rabernat:
@raphaeldussin - is this work in progress, or should we close?

PR #120, opened 2019-03-14
Calculate binary masks from hFac variables

### LINKED ISSUE #44: generate binary masks
Right now there are no explicit mask variables in our xmitgcm datasets.

NEMO produces the following mask variables.
```python
_nemo_keymap_byte_mask = {
    'tmask': 'sea_binary_mask_at_t_location',
    'umask': 'sea_binary_mask_at_u_location',
    'vmask': 'sea_binary_mask_at_v_location',
    'fmask': 'sea_binary_mask_at_f_location',
```

We could create equivalent ones from hFacC, hFacS, and hFacW. This would help with oocgcm compatibility (e.g. lesommer/oocgcm#44).

Question for @serazing: do these variables need to have specific *names*, or does oocgcm look for specific *attributes* (e.g. `standard_name=='sea_binary_mask_at_t_location'`).

timothyas:
This PR adds the 3D mask variables maskC/W/S to grid variables by loading the corresponding hFac variable and computing `np.ceil()` to make 0 or 1. I tried to make the meta data according to conventions discussed in #44, let me know what you think. 

This PR (eventually) resolves #44.

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -544,11 +544,10 @@ class _MDSDataStore(xr.backends.common.AbstractDataStore):
         return vname, dims, data, attrs
 
     def calc_masks(self, vname, data):
-        """Compute mask(C/W/S) as 1 where hFac(C/W/S) nonzero,
-           0 otherwise"""
+        """Compute mask as True where hFac nonzero, otherwise False"""
 
         if vname[0:4] == 'mask':
-            data = np.ceil(data)
+            data = data > 0
 
         return data
 
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -21,7 +21,7 @@ from .variables import dimensions, \
     horizontal_coordinates_curvcart, horizontal_coordinates_llc, \
     vertical_coordinates, horizontal_grid_variables, vertical_grid_variables, \
     volume_grid_variables, state_variables, aliases, package_state_variables, \
-    extra_grid_variables
+    extra_grid_variables, mask_variables
 # would it be better to import mitgcm_variables and then automate the search
 # for variable dictionaries
 
@@ -524,6 +524,9 @@ class _MDSDataStore(xr.backends.common.AbstractDataStore):
                 (vname, dims, data, attrs) = self.fix_inconsistent_variables(
                     vname, dims, data, attrs)
 
+                # Create masks from hFac variables
+                data = self.calc_masks(vname, data)
+
                 thisvar = xr.Variable(dims, data, attrs)
                 self._variables[vname] = thisvar
                 # print(type(data), type(thisvar._data), thisvar._in_memory)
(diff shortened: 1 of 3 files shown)


xmitgcm/mds_store.py
```diff
@@ -540,6 +543,15 @@ def fix_inconsistent_variables(self, vname, dims, data, attrs):
                 data = drc_data
         return vname, dims, data, attrs
 
+    def calc_masks(self, vname, data):
+        """Compute mask(C/W/S) as 1 where hFac(C/W/S) nonzero,
+           0 otherwise"""
+
+        if vname[0:4] == 'mask':
+            data = np.ceil(data)
```
rabernat: we could use a boolean datatype here to save memory
```
data = data > 0
```

Keep in mind that the size of `hFacC` in LLC4320 is 80GB!

rabernat: Just FYI, typecasting works fine here:
```
np.array([False, True]) * np.array([10., 10.])
```
gives `[0., 10.]`

timothyas: Thanks @rabernat! Fixed with bc6258d. 

Test coming soon ...

timothyas: Hey @rabernat, I added a test which ensures that hFac*mask == hFac. This does take a bit of time on the llc grid, for instance, so let me know if you think this is OK... 

Also I'm not sure how the test suite works, is there a main test function that I need to call this function from?  

Thanks!

xmitgcm/test/test_mds_store.py
```diff
@@ -425,19 +425,37 @@ def test_llc_dims(llc_mds_datadirs, method, with_refdate):
...
+    dirname, expected = all_mds_datadirs
+    ds = xmitgcm.open_mdsdataset(
+        dirname, iters=None, read_grid=True,
+        geometry=expected['geometry'])
+
+    hFac_list = ['hFacC', 'hFacW', 'hFacS']
+    mask_list = ['maskC', 'maskW', 'maskS']
+
+    for hFac, mask in zip(hFac_list, mask_list):
+        assert (ds[hFac] * ds[mask] == ds[hFac]).all()
```
rabernat: This looks great. Maybe just use `xr.testing.assert_equal` instead of `.all()`?
http://xarray.pydata.org/en/stable/generated/xarray.testing.assert_equal.html

rabernat:
Sorry about the stickler warnings. It sucks. I'm going to turn it off. Just ignore them.

xmitgcm/test/test_mds_store.py
timothyas: Nice. Well it seems to be about the same computation time, but I'll switch that out if it's preferred

rabernat:
Looks good to me! @raphaeldussin any comments?

raphaeldussin:
LGTM+1 and very useful contribution!

PR #123, opened 2019-03-23
Compatibility with xarray v0.12.0

malmans2:
xmitgcm is not compatible with xarray v0.12.0 because OrderedDict has been removed from `xarray.core.pycompat`. See:
* https://github.com/pydata/xarray/pull/2645/commits/fccb917ef442bbb39ff16a643d700ea52c15934c#
* https://github.com/pydata/xarray/issues/2744

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -34,6 +34,12 @@ if (sys.version_info > (3, 0)):
 else:
     stringtypes = [str, unicode]
 
+# xarray>=0.12.0 compatiblity
+try:
+    from xarray.core.pycompat import OrderedDict
+except ImportError:
+    from collections import OrderedDict
+    
 # should we hard code this?
 LLC_NUM_FACES = 13
 LLC_FACE_DIMNAME = 'face'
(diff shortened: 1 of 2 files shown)


xmitgcm/mds_store.py
```diff
@@ -34,6 +34,12 @@
 else:
     stringtypes = [str, unicode]
 
+# xarray>=0.12.0 compatiblity
+try:
+    from xarray.core.pycompat import OrderedDict
+except ImportError:
+    from collections import OrderedDict
+
```
rabernat: What happens if we just use the one from collections?

rabernat:
Wow good catch! Thanks so much @malmans2!

xmitgcm/mds_store.py
malmans2: I think these are the scenarios:

- It doesn't matter if you don't have cyordereddict installed
- Lower performances if you are using python<3.5. Not sure how much this would impact xmitgcm, but I guess you decided to use the xarray one because of better performances
- Pretty much the same performances if you are using python>=3.5 and xarray<0.12, but xarray uses cyordereddict and xmitgcm collections.

rabernat:
Ok, I'm very happy with this. Thanks @malmans2!

PR #125, opened 2019-03-26
add new CI config

### LINKED ISSUE #124: need to test against multiple version of xarray
As @malmans2 identified in #123, xmitgcm became incompatible with an xarray release.

Since we depend so heavily on xarray, we should probably explicitly test against multiple versions in travis. @raphaeldussin - could you look into modifying our CI configuration to enable this?

raphaeldussin:
tests compatibility with xarray 0.12.0, fixes #124


rabernat:
You will also need to edit .travis.yml.

Now that xarray 0.12 is released, it is the default in the other envs as well. If we want to maintain backwards compatibility, we will have to explicitly specify old xarray versions (e.g. 0.11.3, 0.10.9).

We might also consider adding an env that tests against the latest xarray master (via pip + github). This would have allowed us to catch #123 before the xarray release.

ci/environment-py36-xarray0.12.0.yml
```diff
@@ -0,0 +1,14 @@
+name: test_env
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - python=3.6
+  - xarray>=0.12.0
```
rabernat: Since 0.12.0 is the current release, this environment should be exactly the same as the `py36` environment without a version specifier, right?

What if we added environments for xarray=0.11.3 and 0.10.9? Would this make sense?

raphaeldussin: I am not sure it's been packaged in conda/defaults yet. I had to go through conda-forge to have it installed through conda, default goes to 0.11.3. I think adding tests for older versions make sense.

raphaeldussin: I confirm that v0.12.0 is not yet the default in conda. Right now I only test it in py36 with a specific environment. When it becomes the new default, all python version will test against it.
Also added testing of older versions 0.11.3 and 0.10.9 when feasible and testing of the xarray/master

rabernat:
Fantastic! Thanks for putting all of this together!

A future conversation we should have is when to drop python 2 support.

PR #128, opened 2019-03-27
add binderized examples

raphaeldussin:
what about adding some xmitgcm examples one can run in binder?
Here's a first quick and dirty one that could be the beginning of a more complete set of examples.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 xmitgcm: Read MITgcm mds binary files into xarray
 =================================================
 
-|pypi| |Build Status| |codecov| |docs| |DOI|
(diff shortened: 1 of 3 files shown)


rabernat:
I really like this idea, but I think the binder itself should live in a separate repository from the main code.

raphaeldussin:
submodule for the binder directory?
do you want to put the example notebooks in a submodule as well?

raphaeldussin:
Actually I kind of like example notebooks in the doc section because it builds with RTD.
We would only need to add a binder file to the repo to make it cloud-ready/
@rabernat thoughts on this?

PR #142, opened 2019-05-14
import 2d diags of 3d variables

raphaeldussin:


rabernat:
I will insist that we add a test for this. Here's how I recommend we do it.

- Copy one of the 2D files (like Eta) to a 3D file name (like Theta) and create the appropriate meta file. we should test both built-in state variables and those generated from available_diagnostics.log
- Open those files with `open_mdsdataset` and ensure no errors are raised
- Check that the files have the right dimensions.

raphaeldussin:
breaks a whole lot of tests... we need to get this fixed

timothyas:
Hey @raphaeldussin thank you so much for jumping on this! Sorry I'm late to the party on this - it's been a crazy time preparing. I'll get on this ASAP as well

raphaeldussin:
@timothyas is coming strong with a new PR that should work better

Issue #140, opened 2019-05-07
Unable to read 2D output of 3D fields


timothyas:
Currently `open_mdsdataset` fails when any files which include 2D diagnostics of 3D fields. As an example, the following specification in `data.diagnostics` generates vertically integrated fields:

```
frequency(14) = 2635200.0,
   fields(1:8,14) = 'DFxE_TH ','DFyE_TH ','ADVx_TH ','ADVy_TH ',
                 'DFxE_SLT','DFyE_SLT','ADVx_SLT','ADVy_SLT',
   filename(14) = 'diags/trsp_2d_set1',
#vertically integrate fields we only use to compute vertically integr.
#meridional transports (also omit vertical transports, both to save space)
   fileFlags(14) = ' I      ',
```

However xmitgcm hiccups with this because available_diagnostics.log tells us that these fields, e.g. DFxE_TH, should be 3D but the meta information says they're 2D. 

Hard to think of how to incorporate this because this does not show up in the output meta information via the MITgcm, which I think would be the best solution. Perhaps this could be done by adding the fileFlags field to *.meta whenever it is provided to the diagnostics package. 

So maybe it's not an issue you / we want to fix through xmitgcm (especially because these files could just be read with the low level function `read_3d_llc_data`). Maybe we could add another optional input to open_mdsdataset to handle such conflict, but this might get pretty convoluted...

ifenty:
@raphaeldussin  @rabernat  @timothyas 

We’ve been trying to integrate xmitgcm with our ecco_v4_py package including making tutorials to demonstrate its use for our upcoming ECCO summer school.  We ran into a problem yesterday when trying to load mitgcm data/meta file pairs created using the diagnostics package when the fields are 2D slices of ‘normally’ 3D fields.  For example, if we ask open_mdsdataset to open a data/meta file that contains a 2D slice of THETA (SST) then we get the error
 
  File "/home/ifenty/anaconda2/lib/python2.7/site-packages/xmitgcm/mds_store.py", line 864, in _reshape_for_llc
    assert data.ndim==len(dims)
 
I carefully traced this error back and found that in  load_from_prefix the metadata associated with the variable name is interrogated and contains a field called ‘dims’ which contains, in the case of THETA for example, metadata[‘dims’]=['k', 'j', 'i'].  Later, in _reshape_for_llc there is a check on the array dimension which fails because it expects 4 dimensions (face, k, j, i) while the 2D slice of THETA only contains 3 (face, j, i).    
[...]

PR #143, opened 2019-05-15
extra_metadata takes user input for 2D diags of 3D vars

timothyas:
Title says it all! The only current problem is when 2D and 3D versions of the same variable exist in different files which are read in at the same time. 

See #142 for a list of new tests which need to be made. I would need some guidance in doing that, but I'd be happy to implement them.

### DIFF
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -39,7 +39,8 @@ def test_open_mdsdataset_minimal(all_mds_datadirs):
 
     ds = xmitgcm.open_mdsdataset(
         dirname, iters=None, read_grid=False, swap_dims=False,
-        geometry=expected['geometry'])
+        geometry=expected['geometry'],
+        allow_2d_diags=expected['allow_2d_diags'])
 
     # the expected dimensions of the dataset
     eshape = expected['shape']
@@ -95,7 +96,8 @@ def test_read_grid(all_mds_datadirs):
     dirname, expected = all_mds_datadirs
     ds = xmitgcm.open_mdsdataset(
         dirname, iters=None, read_grid=True,
-        geometry=expected['geometry'])
+        geometry=expected['geometry'],
+        allow_2d_diags=expected['allow_2d_diags'])
 
     for vname in _EXPECTED_GRID_VARS:
         assert vname in ds.variables
@@ -118,12 +120,14 @@ def test_values_and_endianness(all_mds_datadirs):
     # default endianness
     ds = xmitgcm.open_mdsdataset(
                 dirname, iters=None, read_grid=True, swap_dims=False,
-                geometry=expected['geometry'])
+                geometry=expected['geometry'],
+                allow_2d_diags=expected['allow_2d_diags'])
     # now reverse endianness
     ds_le = xmitgcm.open_mdsdataset(
                 dirname, iters=None, read_grid=True, endian='<',
                 swap_dims=False,
-                geometry=expected['geometry'])
+                geometry=expected['geometry'],
+                allow_2d_diags=expected['allow_2d_diags'])
 
     for vname, (idx, val) in expected['expected_values'].items():
         np.testing.assert_allclose(ds[vname].values[idx], val)
@@ -152,7 +156,8 @@ def test_open_dataset_no_meta(all_mds_datadirs):
 
     it = expected['test_iternum']
     kwargs = dict(iters=it, geometry=expected['geometry'], read_grid=False,
-                  swap_dims=False, default_dtype=expected['dtype'])
+                  swap_dims=False, default_dtype=expected['dtype'],
+                  allow_2d_diags=expected['allow_2d_diags'])
 
     # a 3D file
     to_hide = ['T.%010d.meta' % it, 'Eta.%010d.meta' % it]
@@ -167,7 +172,8 @@ def test_open_dataset_no_meta(all_mds_datadirs):
                                             "is not precised (i.e., None)"):
             xmitgcm.open_mdsdataset(dirname, prefix=['T', 'Eta'], iters=it,
                                     geometry=expected['geometry'],
-                                    read_grid=False)
+                                    read_grid=False,
+                                    allow_2d_diags=expected['allow_2d_diags'])
 
     # now get rid of the variables used to infer dimensions
     with hide_file(dirname, 'XC.meta', 'RC.meta'):
@@ -216,7 +222,7 @@ def test_open_dataset_2D_diags(all_mds_datadirs):
 
     it = expected['test_iternum']
     kwargs = dict(iters=it, geometry=expected['geometry'], read_grid=False,
-                  swap_dims=False)
+                  swap_dims=False,allow_2d_diags=True)
 
     to_hide = ['T.%010d.meta' % it, 'T.%010d.data' % it]
     with hide_file(dirname, *to_hide):
@@ -238,13 +244,15 @@ def test_swap_dims(all_mds_datadirs):
     """See if we can swap dimensions."""
 
     dirname, expected = all_mds_datadirs
-    kwargs = dict(iters=None, read_grid=True, geometry=expected['geometry'])
+    kwargs = dict(iters=None, read_grid=True, geometry=expected['geometry'],
+                  allow_2d_diags=expected['allow_2d_diags'])
 
     expected_dims = ['XC', 'XG', 'YC', 'YG', 'Z', 'Zl', 'Zp1', 'Zu']
 
     # make sure we never swap if not reading grid
     assert 'i' in xmitgcm.open_mdsdataset(dirname,
-        iters=None, read_grid=False, geometry=expected['geometry'])
+        iters=None, read_grid=False, geometry=expected['geometry'],
+        allow_2d_diags=expected['allow_2d_diags'])
     if expected['geometry'] in ('llc', 'curvilinear'):
         # make sure swapping is not the default
         ds = xmitgcm.open_mdsdataset(dirname, **kwargs)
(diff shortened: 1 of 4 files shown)


timothyas:
Ok latest commit fixes the issue I mentioned. Now, if 2D diagnostics are recognized then the variable name gets appended with '_2D' within the dataset.

timothyas:
I actually think we could do this without the user input, but it requires an assumption. These lines are the main change 
https://github.com/xgcm/xmitgcm/blob/6dd33bdfe9efdb28a9915c09afc97a97833b9402/xmitgcm/mds_store.py#L686-L697

I noticed that now only lines 690 on are needed. They assume that 3 dims are expected, but only find two (or in the case of LLC which has face dim, it's 4 and 3), then collapse. I'm curious what you think - maybe this assumes too much and so should we still force the user to tell xmitgcm that they expect 2D slices since this is missing from the meta information? 

Thanks guys!

raphaeldussin:
I think this is great! Should we get started with the new tests?

timothyas:
Thanks @raphaeldussin. I'm curious if you think we still need the user to provide the flag '2d_diag_of_3d_var' even though I've found all the tests pass with just the two if statements (without the two involving `extra_metadata`): 

```
         #Expect 3 dims, vertical and 2 horizontal 
         ndims_expected = 3 
         if self.llc: 
             # llc geometry adds face dim 
             ndims_expected += 1 
  
         if len(dims) == 3 and data.ndim == ndims_expected-1: 
             dims = dims[1:] 
             vname = '%s_2D' % vname 
``` 

But like I said I'm worried this would be assuming too much - maybe I'm overthinking things! :P  

As for the tests, I have a rookie question... For reference here is what @rabernat originally suggested: 

- Copy one of the 2D files (like Eta) to a 3D file name (like Theta) and create the appropriate meta file. we should test both built-in state variables and those generated from available_diagnostics.log
- Open those files with open_mdsdataset and ensure no errors are raised
- Check that the files have the right dimensions.

For step 1, do you want to do this all within a test function? So that each time the tests run, it does the steps of copying Eta, making the meta file, writes it out to be loaded in. Or manually copy Eta etc and load it to wherever you guys keep your test data, so that it gets loaded like the other variables each time? If the latter, I'm curious where to load these meta/data files.  Thanks!

rabernat:
I think it would be useful to discuss the optimal design here in general terms, before working more on the implementation. **What do we want to happen?**

Here is the scenario. We have a file, let's say `Theta.data/meta` which, instead of having the expected 3 dimensions, has only 2 dimensions. This can happen if MITgcm diagnostics are configure to output only one k level.

_To preserve backwards compatibility, I would like an error to continue to be raised, as it is now._

However, I would like to add an optional flag like `allow_2d_diags=True` to `open_mdsdataset` which turns on the extra logic @timothyas provided above. After thinking about it more, I agree that `extra_metadata` doesn't have to be involved. 

I can try to tackle the tests. I'll push some proposed tests to your branch later today.

rabernat:
I just added this test:

https://github.com/xgcm/xmitgcm/blob/2fd7ad7a8f8c1077e314aa7a0ee54f1d9dbe6e37/xmitgcm/test/test_mds_store.py#L199-L234

This is not totally complete, but I think it should help @timothyas move forward.
The strange thing is that it passes just fine already with standard grids but doesn't work with LLC grids. I would prefer consistent behavior either way: _either we error when 3D fields are expected or not, regardless of grid type._ I guess the presence of the extra LLC face dimension throws a wrench into the logic somewhere.

xmitgcm/test/test_mds_store.py
```diff
@@ -195,6 +196,44 @@ def test_open_dataset_no_meta(all_mds_datadirs):
...
+        ldir = py.path.local(dirname)
+        old_prefix = 'Eta.%010d' % it
+        new_prefix = 'T.%010d' % it
+        for suffix in ['.data', '.meta']:
+            lp = ldir.join(old_prefix + suffix)
+            lp.copy(ldir.join(new_prefix + suffix))
+
+        # currently raises ONLY with LLC-type datasets, but not regular grids
+        # WHY?!?
+        ds = xmitgcm.open_mdsdataset(dirname, prefix=['T'], **kwargs)
```
rabernat: My question for the group, including @ifenty, is, what sort of API do we want here? Do we want it to "just work" when these 2D diags are encountered, or do we want to raise an error by default? In this case, we would have a new keyword argument like `allow_2d_diags=True` which would turn on the new feature of correctly parsing 2D slices of 3D diags.

timothyas: My opinion is to require the keyword `allow_2d_diags`. Along with this, we can use the if statements above to print out a useful error message. Something like 

```
#Expect 3 dims, vertical and 2 horizontal 
ndims_expected = 3 
if self.llc:
    # llc geometry adds face dim 
    ndims_expected += 1 
  
if len(dims) == 3 and data.ndim == ndims_expected-1:
    if allow_2d_diags: 
        dims = dims[1:] 
        vname = '%s_2D' % vname 
    else: 
        raise TypeError('Variable %s from file prefix %s has unexpected number of dimensions. \n'
                        'Set allow_2d_diags=True to read 2D diagnostic of a typically 3D variable' % (vname,prefix))
```
Perhaps a different error should be used, but this is my general idea. 

I'd also be happy to turn the notebook I've been using for testing into documentation.

timothyas:
Thanks @rabernat for getting me started on the tests!

timothyas:
> I just added this test:
> 
> https://github.com/xgcm/xmitgcm/blob/2fd7ad7a8f8c1077e314aa7a0ee54f1d9dbe6e37/xmitgcm/test/test_mds_store.py#L199-L234
> 
> This is not totally complete, but I think it should help @timothyas move forward.
> The strange thing is that it passes just fine already with standard grids but doesn't work with LLC grids. I would prefer consistent behavior either way: _either we error when 3D fields are expected or not, regardless of grid type._ I guess the presence of the extra LLC face dimension throws a wrench into the logic somewhere.

I think I found why it fails for llc but not others. This snippet lived just a few lines above our little hack in `load_from_prefix`: 
https://github.com/xgcm/xmitgcm/blob/2fd7ad7a8f8c1077e314aa7a0ee54f1d9dbe6e37/xmitgcm/mds_store.py#L670-L679

I think this was to catch output from e.g. barotropic simulations, where we actually have 2 dims, even for grid variables like hFac. I have made a number of changes, I went ahead and implemented what I mentioned earlier: added a keyword `allow_2d_diags`. I then removed the above lines, but then had to add the `allow_2d_diags` to the `expected` dictionaries in the test suite. This should allow us to say "yes we expect 2D diags for the barotropic experiment, but no others".  

However I'm also getting a KeyError: with variable 'T' in get_virtual_variable that I don't quite understand. My hunch is that this has to do with the new test, something to do with the renaming and copying of Eta, but I'm at a loss for that one.

timothyas:
PS perhaps that was assuming of me to just go ahead and implement what I suggested, but I got addicted to solving this. Happy to revert changes if everyone disagrees.

rabernat:
@timothyas - thanks a lot for working on this!

It's not an option to suddenly require a new keyword argument `allow_2d_diags=True` for barotropic simulations that previously worked fine. This would be considered a "breaking change," because it would cause xmitgcm code that previously worked to suddenly stop working if a user upgrades xmitgcm. Imagine the frustration that would cause for you as a user! These sorts of changes must be avoided at all costs. Instead, we should make whatever changes we make here backwards compatible. If we really need to introduce breaking changes, we need to do so very slowly, over many release cycles, starting first with deprecation warnings. I don't think that is necessary here.

I would back up a little bit and rethink the implementation here.

The *default* behavior (`allow_2d_diags=False`) should just leave things as they were before. `allow_2d_diags=True` should turn on a new code path that overcomes the specific problem raised in #140.

> My hunch is that this has to do with the new test, something to do with the renaming and copying of Eta, but I'm at a loss for that one.

Hmm, there could be a problem with my test then. Unfortunately I will not have time to work on this today.

@ifenty - where do we stand on the ECCO summer school. Is there a specific date by which you need this fixed? Have we already missed it? What happens if we fail to implement this new feature by the target date? What work-arounds might exist?

timothyas:
Hey @rabernat, thanks for this thorough explanation. This totally makes sense - of course we don't want to break old code! And it makes sense that you would want these tests to break in order to catch breaking changes like this. I'm learning ... slowly... :) 

So I'm now thinking about what was already in the function 
https://github.com/xgcm/xmitgcm/blob/2fd7ad7a8f8c1077e314aa7a0ee54f1d9dbe6e37/xmitgcm/mds_store.py#L670-L679

and it seems like the default behavior is the equivalent of `allow_2d_diags=True`, but now we're just also trying to catch the LLC case. 

I also don't have a ton of time to work on this right now since I have to prepare some other materials for the school. I'm curious what @ifenty thinks, but I would like to propose the following temporary solution. The students will be using a central distribution of xmitgcm, which is in a community directory on the super computer cloned from git. We could just clone a branch (e.g. from my fork) which has this fix for LLC geometries and roll with that.

ifenty:
@timothyas  

if that's the best we can do before Monday then we'll give the students your xmitgcm branch.

rabernat:
We can easily release a new xmitgcm version over the week.

But I still don't understand...what is the actual fix? Do we know what it is, or it still has to be solved.

I have a hunch it is just a few lines of code.

rabernat:
@timothyas - I made an alternate fix for this in #145.

I really appreciate the effort you put into this. It was definitely not for nothing, as you have learned probably more than you ever wanted to know about the guts of xmitgcm. I think you were on track to reaching the same solution. But you mentioned being pressed for time, so I went ahead with my own solution.

timothyas:
Thanks @rabernat! Yeah I was thinking the same thing, but you wrote it much cleaner, and fixed the rest which I wasn’t sure on. I’m happy enough with the knowledge gained.

PR #149, opened 2019-05-19
Read data with level subsets, read namelists.

aulemahal:
This PR adds a "parse_namelist" function to utils.py that does what you expect: it reads a Fortran namelist as used in the MITgcm into a Python structure (dict of dicts) casting values to the appropriate types.

Also, it modifies the open_mdsdataset function to add the ability to read data that has only a subset of levels, as when the "levels" option is used with pkg/diagnostics. It adds a "levels" kwarg to receive a list of indexes (ints) and sends it to the "_MDSDataStore" so the kranges correspond. However, it reads the grid with the full range of levels and subsets the xarray dataset directly so to have the appropriate coordinates.

CAVEAT:
I don't know what to do with Zp1. Right now, the "deepest" element of everything with the Zp1 coordinate is NaN, since we cannot know what level to use and what to add.

I might have time to add some tests, but for some reasons py.test is having issues on my computer right now...


rabernat:
Hi @aulemahal--thanks a lot for your pull request! I hope to find time to review it soon. We really appreciate your contribution.

rabernat:
Hi @aulemahal - I had a look through your PR and overall I think this is a great contribution!

We definitely need tests for any new features added to xmitgcm. I would recommend the following:
- A unit test for `utils.parse_namelist` which verifies that this function works properly
- An integration test for `open_mdsdataset` which actually reads a dataset that has a namelist (you will need to add one to one of the existing examples)

Let me know how we can help you get py.test up and running in your development environment.

aulemahal:
Hi @rabernat, I was getting on this test development, but I might need help indeed with py.test. When running `py.test` in the root folder of the repo, I get almost no errors, as expected, but I do get some : 
`OSError: [Errno 24] Too many open files: '/tmp/pytest-of-phobos/pytest-0/mdsdata3/global_oce_llc90/T.0000000008.data'`
Also, it seems I can't get to run py.test more than once or twice, there is a memory leak of some sort and my computer just crashes after failing for almost all tests.
Any idea on the matter?

I'm running Python 3.7.3, py.test 4.5.0. (I tested earlier py.test versions, same thing.)
On linux.

rabernat:
Hi @aulemahal - I have discovered the same problem. I will try to look into it.

aulemahal:
So here it is! 
Two new test functions, both using a new dataset that I stole from the verification directory of the MITgcm. I added it through figshare, as the others, but I currently own it, is that the correct way to do it?

rabernat:
@aulemahal - I finally got caught up on xmitgcm and got your PR passing all tests.

I'll merge this unless you have any more comments.

aulemahal:
Oh nice! Thanks!

I say LGTM!

PR #155, opened 2019-07-01
fixed sign of metric terms

cspencerjones:
The variable "metric" wasn't being passed correctly to transform_u_to_v so I was getting the incorrect sign in some of my domain.


rabernat:
Good catch. I'm not sure how I was able to leave this out and have it still work.

In theory we should have tests for this, but it seems like such a huge pain.

Issue #148, opened 2019-05-19
Add parallel option to open_mdsdataset


malmans2:
I think it would be helpful to have the option `parallel` in open_mdsdataset.
I have a dataset with 20k files and this is what I get:

```python
from dask.distributed import Client
import xmitgcm
client = Client(threads_per_worker=1)
print(client)
path = '/home/idies/workspace/OceanCirculation/fromNeil'
for p in [True, False]:
    print('\nparallel={}'.format(p))
    %time xmitgcm.open_mdsdataset(data_dir=path, ignore_unknown_vars=True, parallel=p)
```

    <Client: scheduler='tcp://127.0.0.1:33596' processes=16 cores=16>
    
    parallel=True
    CPU times: user 1min 18s, sys: 17 s, total: 1min 35s
    Wall time: 1min 35s
    
    parallel=False
    CPU times: user 5min 8s, sys: 1min 15s, total: 6min 24s
    Wall time: 5min 48s

I also like a lot being able to see what's going on with the dask dashboard.
All I did is changing this:
https://github.com/xgcm/xmitgcm/blob/310f3c82699f458e24beb1313702a1acf5eda2af/xmitgcm/mds_store.py#L204-L206
with this:
```python
                # wrap the open_mdsdataset with delayed
                if parallel:
                    import dask
                    open_ = dask.delayed(open_mdsdataset)
                else:
                    open_ = open_mdsdataset

                datasets = [open_(data_dir,
                                  iters=iternum,
                                  read_grid=False,
                                  **kwargs) for iternum in iters]
                if parallel:
                    # calling compute here will return the datasets lists,
                    # the underlying datasets will still be stored as dask arrays
                    datasets = dask.compute(datasets)[0]
```
Should I submit a PR?

rabernat:
Hi @malmans2 - thanks for raising this important issue. I agree we really need to speed up the the dataset creation process.

When I have profiled this in the past, I found that nearly all of the time was being spent calling `glob`. I think there are some very simple optimizations we could do that would greatly speed things up. I have been meaning to work on this, and I'm glad you reminded me about it.

I'm curious what exactly is being parallelized in your example above?

malmans2:
I explored the code a little more and indeed I noticed that glob is used in functions that are called in for loops.
For example, in my case (~20k files) about 30s are spent in this checking part:
[...]

PR #162, opened 2019-08-05
Speed up open_mdsdataset using global variables

malmans2:
This PR is trying to speed up `open_mdsdataset` (the discussion started here #148). I tested on a repository with 20k files and the elapsed time dropped from ~4min to ~1min. 

`open_mdsdataset` is calling itself `n=iters` times here: https://github.com/xgcm/xmitgcm/blob/94c7f8ac013a56f3ba4991dc353955ee7001d7c9/xmitgcm/mds_store.py#L204-L206

There are a few functions that are used several times always with the same arguments (most of them use glob in some way). I timed different sections of the code, and it looks like the slowest are:

- https://github.com/xgcm/xmitgcm/blob/94c7f8ac013a56f3ba4991dc353955ee7001d7c9/xmitgcm/mds_store.py#L174-L181
- https://github.com/xgcm/xmitgcm/blob/94c7f8ac013a56f3ba4991dc353955ee7001d7c9/xmitgcm/mds_store.py#L501-L522

Here is what I'm doing to speed up ([here](https://github.com/malmans2/xmitgcm/blob/9078dcc84928c2e49d5dddbab8a40f6285d3c298/xmitgcm/mds_store.py#L776-L817) is an example):
1) Use `inspect` in the slowest functions to store frames.
2) Frames and their output are stored as global variables.
3) Compare the current frame with the same function previously executed.
4) If current and previous frames used the same arguments, then return the output of the previous one without having to run the whole function again.

A few comments:
1) I want to make sure that the global variables are only used if open_mdsdataset is calling itself. That's why I'm popping global variables [here](https://github.com/malmans2/xmitgcm/blob/9078dcc84928c2e49d5dddbab8a40f6285d3c298/xmitgcm/mds_store.py#L143-L149)
2) The slowdown of `_get_all_matching_prefixes` is due to `glob` (and fnmatch, which I think is internally used by `glob`). So I'm reading just once using `listdir`, then I'm using a for loop combined with `.endswith` rather than `fnmatch` (`fnmatch` is quite slower, and this specific function only cares about `*.iter.ext`). See https://github.com/xgcm/xmitgcm/issues/148#issuecomment-501408825
3) I've only applied this approach to the functions that seem to be slower, but there are more functions that could be edited in the same way. 
4) It looks like all tests related to these edits are successful on my machines. I didn't make additional tests, so let me know if you can think about cases where problems due to these edits would not be detected by the current tests.

We didn't really discuss the use of global variables, so I'm not sure if this is the best approach. Let me know your thoughts! Also, I'm adding comment about the failing tests not related to this PR in separate issues. If it turns out that this PR is the way to go, let me know if you want to address them here.

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -39,7 +39,7 @@ try:
     from xarray.core.pycompat import OrderedDict
 except ImportError:
     from collections import OrderedDict
-    
+
 # should we hard code this?
 LLC_NUM_FACES = 13
 LLC_FACE_DIMNAME = 'face'
@@ -139,6 +139,14 @@ def open_mdsdataset(data_dir, grid_dir=None,
     _, _, _, arg_values = inspect.getargvalues(frame)
     del arg_values['frame']
     function_name = inspect.getframeinfo(frame)[2]
+
+    # if open_mdsdataset did not call itself,
+    # clear global variables created to speed up
+    if inspect.getframeinfo(frame)[0] != inspect.getframeinfo(frame.f_back)[0]:
+        glob_vars = ['_prev_layers', '_prev_all_grid_variables',
+                     '_prev_all_data_variables', '_prev_all_matching_prefixes']
+        for var in glob_vars:
+            globals().pop(var, None)
 
     # auto-detect whether to swap dims
     if swap_dims is None:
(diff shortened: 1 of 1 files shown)


rabernat:
Sorry it has taken me so long to look at this. My month has just been slammed with travel.

Everything you described seems reasonable, *except* the global variables part. Global variables are usually the sign of a poor design. Can we think of another way to do it?

rabernat:
What if we used a cache instead?

https://www.thepythoncorner.com/2018/04/how-to-make-your-code-faster-by-using-a-cache-in-python

rabernat:
I just tried an alternative approach in #168.

PR #165, opened 2019-08-06
Make dask and mmap mutually exclusive

malmans2:
Closes #156

PR #167, opened 2019-08-15
Fix deprecation errors in tests.

aulemahal:
This PR replaces the deprecated `pytest.raises(..., message="...")` by the pytest-suggested recipe.
Fixes #167 .

[Link to the pytest docs.](https://docs.pytest.org/en/4.6-maintenance/deprecations.html#message-parameter-of-pytest-raises)

This is the reason #165 still fails. 
I hope that with these two PRs, tests will become useful again!

### DIFF
--- a/xmitgcm/test/test_mds_store.py
+++ b/xmitgcm/test/test_mds_store.py
@@ -163,11 +163,12 @@ def test_open_dataset_no_meta(all_mds_datadirs):
         assert ds['Eta'].dims == dims_2d
         assert ds['Eta'].values.ndim == len(dims_2d)
 
-        with pytest.raises(IOError, message="Expecting IOError when default_dtype "
-                                            "is not precised (i.e., None)"):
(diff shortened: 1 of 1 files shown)


rabernat:
@aulemahal - thanks for this important fix! I am going to try restarting the travis build, which had some weird errors.

Do you know if your fix is python 2 compatible?

rabernat:
It looks like a regression in fsspec (https://github.com/intake/filesystem_spec/issues/123) has broken llcreader. The current test failures are unrelated to this PR.

aulemahal:
This should be python 2 compatible. The Deprecation Error was introduced in pytest 4.1 and pytest 4.6 is the latest Python 2 compatible version. Also, let's not forget Python 2's EOL is in 4 months!

rabernat:
The failures here are related to https://github.com/intake/filesystem_spec/issues/123. This PR is good to go.

PR #168, opened 2019-08-30
Use cache for listing directories

### LINKED ISSUE #148: Add parallel option to open_mdsdataset

rabernat:
Hopefully fixes #148. (Alternative to #162)

All the directory listing and matching is placed in the `file_utils.py` modules. These functions use cachetools to cache their output.

I have not actually tested this on a large archive yet. So far I'm just trying to get the tests to pass. I think I have uncovered some weird properties of the tar archives we are using for test data. I want to see what happens on travis.

### DIFF
--- a/ci/environment-py27-xarray0.10.9.yml
+++ b/ci/environment-py27-xarray0.10.9.yml
@@ -6,7 +6,7 @@ dependencies:
   - numpy
   - pytest
   - future
-  - cachetools
   - pip:
+    - cachetools
     - codecov
     - pytest-cov
--- a/ci/environment-py27-xarray0.10.9.yml
+++ b/ci/environment-py27-xarray0.10.9.yml
@@ -6,6 +6,7 @@ dependencies:
   - numpy
   - pytest
   - future
+  - cachetools
   - pip:
     - codecov
     - pytest-cov
--- a/ci/environment-py27-xarray0.11.3.yml
+++ b/ci/environment-py27-xarray0.11.3.yml
@@ -6,7 +6,7 @@ dependencies:
   - numpy
   - pytest
   - future
-  - cachetools
   - pip:
+    - cachetools
     - codecov
     - pytest-cov
--- a/ci/environment-py27-xarray0.11.3.yml
+++ b/ci/environment-py27-xarray0.11.3.yml
@@ -6,6 +6,7 @@ dependencies:
   - numpy
   - pytest
   - future
+  - cachetools
   - pip:
     - codecov
     - pytest-cov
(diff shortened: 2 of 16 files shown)


rabernat:
In `test_multiple_iters`, I am seeing these kinds of errors

```
E                           OSError: Could not find the expected file prefixes ['._V', '._U', '._W', 'S', '._S', '._Eta', '._T', 'W', 'Eta', 'T', 'V', 'U'] at iternum 100. (Instead found ['._U', 'PHL', '._V', '._W', '._PHL', 'S', '._Eta', '._PH', '._S', '._T', 'W', 'Eta', 'T', 'PH', 'V', 'U'])
```

This reveals that our test data actual have these weird ghost files in them! This seems likely to be the culprit:
https://superuser.com/questions/259703/get-mac-tar-to-stop-putting-filenames-in-tar-archives

The unfortunate conclusion is that we need to regenerate the test data. 🤢

malmans2:
Oh right, I've seen this before and I just forced `listdir` to ignore hidden files as `globus`.
A `for` loop and `.startswith('.')` shouldn't be to expensive now that you `listdir` only once.
But if you think that there are cases where users care about hidden files, than it's definitely the wrong approach.

PR #172, opened 2019-09-27
rename repo to MITgcm/xmitgcm

rabernat:
I think this gets all of the relevant github urls.


rabernat:
Travis worked too, so we should be good to go!

PR #194, opened 2020-03-04
Python 3.8 compatibility

roxyboy:
Updated some functions that had deprecation warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,10 @@ matrix:
     env: CONDA_ENV=py36
   - python: 3.6
     env: CONDA_ENV=py36-xarraymaster
+  - python: 3.7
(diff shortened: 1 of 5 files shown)


roxyboy:
The tests all pass in Python 3.6-3.8 but fails in Python 2.7 due to the function `dask.drop_var()` not being supported. How do you think about dropping support for Python 2.7 @rabernat ??

rabernat:
I think we can drop python 2 in future releases. Thanks for this!

roxyboy:
Ok, all tests in py27 and py3 pass now :)

rabernat:
I guess this will also fix the annoying deprecations warnings from xarray.auto_combine. Great work Takaya!

PR #195, opened 2020-03-12
Add a function to get the full-size of xg and yg from input files

cspencerjones:
This function gets XG and YG from .mitgrid files. They are the full size (i.e. they include the points on the right hand side of the final gridpoint in both x and y directions. This is useful if you want to determine which tile and individual lat/lon point is on.

### DIFF
--- a/xmitgcm/test/test_utils.py
+++ b/xmitgcm/test/test_utils.py
@@ -973,6 +973,49 @@ def test_get_grid_from_input(all_grid_datadirs, usedask):
                                      dtype=np.dtype('d'), endian='>',
                                      use_dask=False,
                                      extra_metadata=None)
+            
+            
+            
+@pytest.mark.parametrize("usedask", [True, False])
+def test_get_xg_yg_from_input(all_grid_datadirs, usedask):
+    from xmitgcm.utils import get_xg_yg_from_input, get_extra_metadata
+    from xmitgcm.utils import read_raw_data
+    dirname, expected = all_grid_datadirs
+    md = get_extra_metadata(domain=expected['domain'], nx=expected['nx'])
+    tx=30
+    ty=30
+    bl=[1,2,3]
+    ds = get_xg_yg_from_input(dirname + '/' + expected['gridfile'],
+                             geometry=expected['geometry'],
+                             dtype=np.dtype('d'), endian='>',
+                             use_dask=usedask,
+                             extra_metadata=md,
+                             tilex=tx,tiley=ty,
+                             blankList=bl)
+    # test types
+    assert type(ds) == xarray.Dataset
+    assert type(ds['XG']) == xarray.core.dataarray.DataArray
+
+    if usedask:
+        ds.load()
+
+    # check all variables are in
+    expected_variables = ['XG', 'YG']
+
+    for var in expected_variables:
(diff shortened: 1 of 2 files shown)


raphaeldussin:
what was wrong with get_grid_from_input ?

https://github.com/MITgcm/xmitgcm/blob/1fb9cc2520d0df1cf199b347d94c91853d22dddb/xmitgcm/utils.py#L1247

cspencerjones:
It doesn't do what I want - it doesn't get the rightmost gridpoints.

xmitgcm/test/test_utils.py
```diff
@@ -973,6 +973,49 @@ def test_get_grid_from_input(all_grid_datadirs, usedask):
...
+    ds = get_xg_yg_from_input(dirname + '/' + expected['gridfile'],
+                             geometry=expected['geometry'],
+                             dtype=np.dtype('d'), endian='>',
+                             use_dask=usedask,
+                             extra_metadata=md,
+                             tilex=tx,tiley=ty,
+                             blankList=bl)
+    # test types
+    assert type(ds) == xarray.Dataset
+    assert type(ds['XG']) == xarray.core.dataarray.DataArray
```
rabernat: use `isinstance` here

rabernat (changes requested):
Thanks for looking into this @cspencerjones. I appreciate that the existing `get_grid_from_inputs` function did not work for the reasons you stated. However, I would strongly prefer that we modify that function to meet your needs, rather than introduce a new one with much of the same code.

I see two separate features here:
- The ability to get the extra "outer" grid point from the mitgrid files. This could be an optional flag to `get_grid_from_inputs`, maybe `outer=False` as a default, to keep backwards compatibility
- The ability to reshape a dataset in tiles of a specified shape.

Could we do these as two separate functions, or is there some fundamental reason why they need to be in a single function?

rabernat:
@cspencerjones - are you planning to continue to work on this?

cspencerjones:
It's not my priority. If we think that these changes would be useful for other people then I can but I'm not sure that there is actually need for them right now. So probably no, I will just keep using my own branch. Do you have a strong opinion?

AaronDavidSchneider:
Hi @cspencerjones,
I could really use the outer grid points (xESMF needs outer grid points for conservative regridding). I would be happy if this PR could be somehow resumed :D I'd be happy to help.

cspencerjones:
@AaronDavidSchneider I do not have time to work on this at the moment, since my project has moved away from this problem, at least temporarily. The code is quite dense and hard to follow, and as @rabernat mentions above my new code actually repeats a lot of what's in the original function. However you are welcome to take it and see how it works for you. If it does, then maybe you could contribute to this pull request as well?

rabernat:
Yes it would be great if @AaronDavidSchneider would be able to pick up this PR! We will be happy to help you.

From a technical standpoint, it is probably easiest for you to check out @cspencerjones's branch and make a new PR from your own fork.

AaronDavidSchneider:
> From a technical standpoint, it is probably easiest for you to check out @cspencerjones's branch and make a new PR from your own fork.

That sounds good. I will try my best to get it working. I am not quite sure when I will have the time for it.

PR #208, opened 2020-06-17
add xg and yg to llcreader known_models

### LINKED ISSUE #207: Regarding the LLC4320 grid variables
For our research purposes we want to save surface velocity output from the LLC4320 of only a small region (at the equator) to our local computer as netcdf, preferably using the lon/lat coordinates (as in XG, YG, XC, YC). However, using: 
`ds = model.get_dataset(varnames=['U','V'], iter_start=time_index, iter_stop=time_index+1, k_levels=[0], read_grid = True)`
I only seem to get access to the XC and YC coordinates. How should I interpret the coordinates of U and V (i, j, i_g and j_g)? Can I create the XG and YG coordinates myself (assuming a C-grid), or do I need to do something else?

rabernat:
Fixes #207.

This also unearthed a bug that allowed me to remove some unused code paths.

### DIFF
--- a/xmitgcm/llcreader/known_models.py
+++ b/xmitgcm/llcreader/known_models.py
@@ -45,7 +45,7 @@ class LLC2160Model(BaseLLCModel):
                      'hFacC','hFacS','hFacW','PHrefC','PHrefF',
                      'RAC','RAS','RAW',
                      'RC','RF','RhoRef','rLowC','rLowS','rLowW',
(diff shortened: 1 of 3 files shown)


rabernat:
While we're at it, are there any other missing grid variables?

timothyas:
Nice! There are also the files `DXV`, `DYU`, & `RAZ` which are on the portal for both LLC2160 and 4320. I had commented those out because I didn't see a fix to the "corner point" variables.

PR #214, opened 2020-07-16
Including data from the Arctic cap facet to transformed data when using llcrearrange

### LINKED ISSUE #199: llcreader & LLC grids: Transformations and inclusion the arctic face
This is an issue, but also a notice on something I have been working that will address a couple of issues I have been having with my particular project that I think will be helpful to the community.

**1)** Interpolate/differentiate across faces all points in the dataset (not completely implemented on xgcm when faces appears as a dimension).
**2)** Reduce the amount of NaNs when creating a cutout of the dataset when specifying a range of lats and lons.
**3)** Incorporate the Arctic face on the new dataset rather than work separately.

Before, when specifying a range of lats and lons, masking everything outside and dropping as many nans as possible, the result is a new dataset with few faces (renamed), and keeps the associated face connections. This is shown in figure below (SST on llc4320). Not ideal for 4D variables.

cutout_faces_2_6_10

My code tackles the issues above by transforming the dataset into a new one without face as a dimension while preserving/not manipulating the original grid - at least to the best extent. You can apply this transformation to all 12 faces, or a subset as long as these connect (e.g. face 2, 6 and 10). By gerring rid of faces, I recover ``xgcm``'s functionality, plus it incorporates the Arctic. The transformation works well, and it seems it would fit within a class on the LLC reader, or something like that (after communicating with @rabernat). The code so far creates a rectangular grid on logical space and includes the arctic as a series of crowns extending all over the logical domain (or a subset, below I show the whole domain). 

In the figure below, ``Depth`` has been interpolated to vorticity (G)-points, and used for masking the zeta/f, as well as plotting it in the background (black colors)

LLC90_vorticity

Creating the arctic crown and combing all faces is equivalent to rotations of 90 degrees (or a multiple of 90) on logical space (U is locally aligned with i, V with j) on some faces, so that transforming vector fields results from renaming the variables (e.g. U -> V, or so), and similarly for the grid and grid metrics. Some careful test needs to be done on the grid and grid metrics- so far looks good, so that they are continuous across all the domain. So far, things are looking pretty good with the grid (some little bugs which wont take much to address), plus make it compatible within llcreader...

Now, specifying a range of lat + lons I can create funny looking cutouts that contain the least amount of NaNs. 
LLC4320_SST_cutout

* Has somebody worked on a similar approach (i.e. conflicting work)?
* Where would this have a better fit within xmitgcm? @rabernat suggested llcreader...
* Also, the code so far is able to transform and manipulate (i.e. apply diff/interp), without visible discontinuities across old face boundaries, all variables on the ECCO data. But the new dataset does split the Arctic (not ideal for people focusing on the Arctic). An alternative is to create several transformations, a possibility shown above, and another where the dataset is centered around the Arctic, and so on...

Mikejmnez:
Here is an initial pull request of my code which facilitates including data from the arctic cap (facet) into a transformed array which no longer has 'face' as a dimension. Before, data was transformed into a single (rectangular) array with data from the arctic cap missing. Now, arctic is included as a form of 'crown', which is done by splitting the arctic data into four different (triangular) regions and concatenating each triangular region with the corresponding facet with which these 'exchange' data. Below is an example of LLC4320 data transformed using my xmitgcm fork

arctic_crown_LLC4320

This pull request should close #199 .

I must mention that I have a more general code that allows for transformation of a subset of faces (as long as these are contiguous) into a single array with face no longer a dimension, as opposed to all the faces. This allows for an economic version of the transformation since not all data is needed, nor transformed. In addition, I have also created a transformation that is centered around the arctic, creating an array that resembles a polar projection centered on the north pole. This is in my repository https://github.com/Mikejmnez/llc_transformations, and there you can find an example notebook. I would be interested in including such code in ```llcreader```, but would take some time to figure out the best way to do so, and of course some guidance.

cheers!
Miguel

### DIFF
--- a/xmitgcm/llcreader/llcmodel.py
+++ b/xmitgcm/llcreader/llcmodel.py
@@ -197,10 +197,6 @@ def _arct_crown(ds, varName, metrics=['dxC', 'dyC', 'dxG', 'dyG']):
         arct_facet: New arctic cap data into a new facet, associated with
                     varName
     '''
-    if 'tile' in ds.dims:
-        face = 'tile'
-    else:
-        face = 'face'
     arc_cap = 6
     ARCT = []
     afaces = [2, 5, 7, 10]  # order of faces with which artic cap connects
@@ -208,7 +204,7 @@ def _arct_crown(ds, varName, metrics=['dxC', 'dyC', 'dxG', 'dyG']):
         if k == 2:
             fac = 1
             _varName = varName  # copy, b/c varName may change
-            DIMS = [dim for dim in ds[_varName].dims if dim != face]
+            DIMS = [dim for dim in ds[_varName].dims if dim != 'face']
             dims = Dims(DIMS[::-1])
             dtr = list(dims)[::-1]
             dtr[-1], dtr[-2] = dtr[-2], dtr[-1]
@@ -219,7 +215,7 @@ def _arct_crown(ds, varName, metrics=['dxC', 'dyC', 'dxG', 'dyG']):
             y0, yf = 0, int(len(ds[dims.X]))
             xslice = slice(x0, xf)
             yslice = slice(y0, yf)
-            da_arg = {face: arc_cap, dims.X: xslice, dims.Y: yslice}
+            da_arg = {'face': arc_cap, dims.X: xslice, dims.Y: yslice}
             sort_arg = {'variables': dims.Y, 'ascending': False}
             mask_arg = {dims.X: xslice, dims.Y: yslice}
             if len(dims.X) + len(dims.Y) == 4:  # vector field or metric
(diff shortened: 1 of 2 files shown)


rabernat:
Thanks for contributing this @Mikejmnez! It looks like a great start.

I can see you are still working on things. Let me know when you're ready for a review.

Mikejmnez:
> Thanks for contributing this @Mikejmnez! It looks like a great start.
> 
> I can see you are still working on things. Let me know when you're ready for a review.

@rabernat  Yeah, just going through the tests ( I hadn't had time until now for this). But for now, I'd be interested in your opinion when it comes to potentially including the rest of my code that works with faces on ```llcreader``` - if it makes sense.  I might have shown some of this to you on a previous `Poseidon` meeting: Transform only a subset of faces as long as these connect (e.g. remove ```face``` as a dimension) as opposed to all the faces. For example, provide the subset of faces ```faces={2, 6, 10}``` and therefore the transformation (and new dataset) only takes into account these. If ```faces={0,2,10}```, return error since these don't connect. 

Also, a new transformation that centers the new data on the arctic cap (only transforms faces 2, 5, 6, 7 and 10), although there are some nans as a result.

Code is in https://github.com/Mikejmnez/llc_transformations , with a notebook with some examples with surface llc4320 data.

rabernat:
Just following up on this issue. In generally, I am happy to accept these new features into xgcm as long as they
- [ ] are covered by tests
- [ ] do not require deep changes to the existing code base
- [ ] follow the same coding style and principles as the rest of this package (e.g. delayed evaluation)

Your code is on track to meet these expectations. So 🚀 !

PR #223, opened 2020-09-18
Travis to gha

timothyas:
Migrate to GitHub Actions, just following up on my promise! 

One issue I ran into, GHA for some reason had trouble with the `test_listdir` test. Sorting the lists of files before comparing removed the issue. Seems innocuous.

Two questions:
1. `aiohttp` had to be added to the yaml environment files, which I guess is used by fsspec. Should this be added to dependencies in setup.py? 

2. With a codecov.yml file, do we need .coveragerc?

Thanks to @jbusecke and others for leading the charge at xgcm!


jbusecke:
> With a codecov.yml file, do we need .coveragerc?

We have consolidated those over at [xgcm](https://github.com/xgcm/xgcm/pull/233)

timothyas:
Thanks @jbusecke!

rabernat:
Sorry for not noticing this earlier! Looks great.

timothyas:
No problem!

Issue #224, opened 2020-09-25
xmitgcm dev branch: possible bug/broken dependency


urielz:
First off, thank you devs for this extremely useful library.

I have 0.4.1 running fine on my main workstation. I came across this issue while trying to install xmitgcm dev branch from scratch on a new machine. If a run this simple code:

```
import numpy as np
import xarray as xr
from xmitgcm import llcreader
import os

model = llcreader.ECCOPortalLLC4320Model()

fprefix = 'Eta'
outdir = '.'

klev = 1
for ii in range(0,klev):

    ds = model.get_dataset(varnames=[fprefix],k_levels=[ii])
    region_slice = {'face': slice(10,11)}
    region = ds.isel(**region_slice,k=0,time=0)
    iters, datasets = zip(*region.groupby('k_l'))


    fname = os.path.join(outdir, '%s.%03d.nc' % (fprefix, ii))
    xr.save_mfdataset(datasets, [fname], engine='netcdf4')
```

I get this error:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-2-1d330d174849> in <module>
     20     fname = os.path.join(outdir, '%s.%03d.nc' % (fprefix, ii))
     21     print(fname)
---> 22     xr.save_mfdataset(datasets, [fname], engine='netcdf4')
     23 
     24 

~/.local/share/virtualenvs/LLC_d2-j1O87tS3/lib/python3.8/site-packages/xarray/backends/api.py in save_mfdataset(datasets, paths, mode, format, groups, engine, compute)
   1237 
   1238     try:
-> 1239         writes = [w.sync(compute=compute) for w in writers]
   1240     finally:
   1241         if compute:

~/.local/share/virtualenvs/LLC_d2-j1O87tS3/lib/python3.8/site-packages/xarray/backends/api.py in <listcomp>(.0)
   1237 
   1238     try:
-> 1239         writes = [w.sync(compute=compute) for w in writers]
   1240     finally:
   1241         if compute:

~/.local/share/virtualenvs/LLC_d2-j1O87tS3/lib/python3.8/site-packages/xarray/backends/common.py in sync(self, compute)
    153             # targets = [dask.delayed(t) for t in self.targets]
    154 
--> 155             delayed_store = da.store(
    156                 self.sources,
    157                 self.targets,

[...]

PR #225, opened 2020-09-28
Fix llcreader klevels bug

rabernat:
Possible fix for #224.

Will travis even run on this repo any more? It seems broken in all our other repos.

### DIFF
--- a/xmitgcm/llcreader/llcmodel.py
+++ b/xmitgcm/llcreader/llcmodel.py
@@ -450,7 +450,7 @@ def _get_1d_chunk(store, varname, klevels, nz, dtype):
     buffer = file.read(read_length)
     data = np.frombuffer(buffer,dtype=dtype)
 
-    # now subset
+    # now subset: is this line broken?
     return data[klevels]
 
 class BaseLLCModel:
--- a/xmitgcm/llcreader/llcmodel.py
+++ b/xmitgcm/llcreader/llcmodel.py
@@ -450,7 +450,6 @@ def _get_1d_chunk(store, varname, klevels, nz, dtype):
     buffer = file.read(read_length)
     data = np.frombuffer(buffer,dtype=dtype)
 
-    # now subset: is this line broken?
     return data[klevels]
 
 class BaseLLCModel:
@@ -533,12 +532,12 @@ class BaseLLCModel:
         # determine kp1 levels
         # get borders to all k (center) levels
         # ki used to get Zu, Zl later
-        ku = np.concatenate([k_levels[1:],[k_levels[-1]+1]])
+        ku = k_levels[1:] + [k_levels[-1] + 1 ]
         kp1 = []
         ki=[]
         for i,(x,y) in enumerate(zip(k_levels,ku)):
-            kp1+= [x] if x not in kp1 else []
-            kp1+= [y] if y-x==1 else [x+1]
+            kp1 += [x] if x not in kp1 else []
+            kp1 += [y] if y-x==1 else [x+1]
 
 
         kp1=np.array(kp1)
@@ -711,7 +710,7 @@ class BaseLLCModel:
         if type=='latlon':
             ds = _faces_coords_to_latlon(ds)
 
-        k_levels = k_levels or np.arange(self.nz)
+        k_levels = k_levels or list(range(self.nz))
         kp1_levels = self._get_kp1_levels(k_levels)
 
         ds = ds.sel(k=k_levels, k_l=k_levels, k_u=k_levels, k_p1=kp1_levels)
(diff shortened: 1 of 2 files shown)


timothyas:
I think travis is not working here just because the fsspec seems to require aiohttp, which needs to be in the conda environment. I had to add this to the environment files in #223 e.g. [here](https://github.com/MITgcm/xmitgcm/blob/3698662aa4a94ba775aea9438fa4313a6449064e/ci/environment-py3.7.yml)

rabernat:
This did not immediately reproduce the error reported in #224 (although it did catch a failure for python 2.7). We will have to dig deeper.

xmitgcm/test/test_llcreader.py
```diff
@@ -124,3 +127,17 @@ def test_ecco_portal_latlon(ecco_portal_model):
...
+    # just get three timesteps
+    iter_stop = ecco_portal_model.iter_start + ecco_portal_model.iter_step + 1
+    ds_vel = ecco_portal_model.get_dataset(
+        varnames=['U', 'V'],
+        type='latlon',
+        k_levels=[1],
+        iter_stop=iter_stop,
+        read_grid=False
+    )
+    ds_vel.sum().compute()
```
rabernat, line 143: @fraserwg and @urielz: I have added the above test to check for the errors you reported in #224 and #233. The test is passing even with the latest master version of xarray.

I was wondering if you can help me figure out what is going on here. _Could you try running the test suite in your local environment?_ You will have to clone my branch. For example:

```bash
git clone -b fix-llcreader-klevels-bug https://github.com/rabernat/xmitgcm.git
cd xmitgcm
pip install -e .
py.test -v
```

fraserwg:
I've tried running your test suite in my environment and everything passes. I've found that the error seems to be a result of a squeeze operation which I omitted from my initial bug report.

The following (modified) test function, however, fails:

```python
@pytest.mark.slow
def test_llc4320_klevels_bug(ecco_portal_model):
    # just get three timesteps
    iter_stop = ecco_portal_model.iter_start + ecco_portal_model.iter_step + 1
    ds_vel = ecco_portal_model.get_dataset(
        varnames=['U', 'V'],
        type='latlon',
        k_levels=[1],
        iter_stop=iter_stop,
        read_grid=True
    )
    ds_vel.squeeze().sum().compute()
```

It does not fail if `read_grid=False`, suggesting the behaviour may still be inconsistent?

rabernat:
> It does not fail if `read_grid=False`, suggesting the behaviour may still be inconsistent?

Ok this is a useful clue. Thanks for checking.

xmitgcm/llcreader/llcmodel.py
```diff
@@ -533,12 +533,12 @@ def _get_kp1_levels(self,k_levels):
         # determine kp1 levels
         # get borders to all k (center) levels
         # ki used to get Zu, Zl later
-        ku = np.concatenate([k_levels[1:],[k_levels[-1]+1]])
+        ku = k_levels[1:] + [k_levels[-1] + 1 ]
```
rabernat, line 536: Note, the original code was producing a float array (the cause of #224). Avoiding numpy ensures everything is int.

rabernat:
By adding a more comprehensive test, I discovered a couple of new bugs. So I ended up refactoring some of the internal logic.

xmitgcm/llcreader/llcmodel.py
timothyas: Nice, that is pretty straightforward

xmitgcm/llcreader/llcmodel.py
```diff
@@ -590,20 +590,24 @@ def _dask_array(self, nfacet, varname, iters, klevels, k_chunksize):
...
+        def _key_and_task(n_k, these_klevels, n_iter=None, iternum=None):
+            if n_iter is None:
+                key = name, n_k, 0, 0, 0
+            else:
+                key = name, n_iter, n_k, 0, 0, 0
+            task = (_get_facet_chunk, self.store, varname, iternum,
+                     nfacet, these_klevels, self.nx, self.nz, self.dtype,
+                     self.mask_override)
+            return key, task
+
```
timothyas, line 602: so clean...  👍

timothyas (approved):
This looks good to me @rabernat, thanks!

xmitgcm/llcreader/llcmodel.py
```diff
@@ -436,7 +436,8 @@ def _get_facet_chunk(store, varname, iternum, nfacet, klevels, nx, nz, dtype,
         data.shape = facet_shape
         level_data.append(data)
 
-    return np.concatenate(level_data, axis=1)
+    out = np.concatenate(level_data, axis=-4)
```
timothyas, line 439: Was this the line causing "extra bugs?

rabernat: Yes. We were getting errors about grid variables having the wrong shape on computation. It's because the concatenation axis is supposed to be the vertical axis. This line assumed the vertical axis was axis 1, but that is not the case if the array has no time dimensions.

timothyas: Makes sense to me, nice catch!

PR #226, opened 2020-10-07
python 2.7 env with old zarr

timothyas:
Current python 2.7 tests break because of zarr, which has not supported python 2 since version 2.3.2. This PR specifies `zarr<=2.3.2` in the ci/environment file.

### DIFF
--- a/ci/environment-py2.7.yml
+++ b/ci/environment-py2.7.yml
@@ -7,7 +7,7 @@ dependencies:
   - pytest
   - future
   - cftime
-  - zarr
+  - zarr<=2.3.2
   - pip:
     - cachetools
     - codecov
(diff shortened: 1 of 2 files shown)


rabernat:
Thanks Tim! Can we just drop python 2 going forward for xmitgcm? Or maybe make a release now and then drop on the next release?

timothyas:
Hmm, I'm not sure what is best so I'll leave that to you. Dropping python 2 would be fine with me, would that just involve removing python 2.7 tests and dropping it from setup? Happy to move forward with either.

rabernat:
Let's merge this, release 0.5.0, and then drop python 3 for 0.6.

PR #230, opened 2020-10-15
Lazy open_mdsdataset with recent xarray versions.

### LINKED ISSUE #229: Lazy `open_mdsdataset`
With xmitgcm 0.4.1 and xarray 0.16.2, it appears that `open_mdsdataset` is not _lazy_, it loads data somewhere in the process, which can use huge amounts of memory with large datasets.
The culprit is `xr.combine_by_coords` here: 
https://github.com/MITgcm/xmitgcm/blob/a5d297978ab43326e933548e18735d2db9aabd3a/xmitgcm/mds_store.py#L241
If I understand correctly, xarray receives variables without indexes and must resolve to loading some data to make the merge. Changing the line to: 
```python3
ds = xr.combine_by_coords(datasets, compat='override', coords='minimal', combine_attrs='drop')
```
I had seemingly the same output, but with no data loading, memory use stayed low, and opening the file was much faster. 

However, I'm guessing this might not be a general fix? My `open_mdsdataset` looks like:
```python3
ds = open_mdsdataset('output/folder/', prefix='dynDiag', delta_t=1, geometry='cartesian', ignore_unknown_vars=True)
```
Data was generated using the `diagnostics` package.

I would have time for a PR, but I don't know xmitgcm enough to see all possible edge cases here!

aulemahal:
Fixes #229 by skipping conflict checks when merging datasets.

I'm not 100% sure this does not break something, but it didn't break the tests on my machine!

### DIFF
--- a/xmitgcm/mds_store.py
+++ b/xmitgcm/mds_store.py
@@ -238,7 +238,7 @@ def open_mdsdataset(data_dir, grid_dir=None,
                 elif xr.__version__ < '0.15.2':
                     ds = xr.combine_by_coords(datasets)
                 else:
-                    ds = xr.combine_by_coords(datasets,combine_attrs='drop')
+                    ds = xr.combine_by_coords(datasets, compat='override', coords='minimal', combine_attrs='drop')
 
                 if swap_dims:
                     ds = _swap_dimensions(ds, geometry)


rabernat:
Sorry for letting this hang for so long. This looks perfect.

PR #236, opened 2020-11-11
fixed probable typo for 'coordinate' entry in dimensions dictionary

ifenty:
convention here seems to be

    XC=dict(dims=["j", "i"], attrs=dict(
        standard_name="longitude", long_name="longitude",
        units="degrees_east", coordinate="YC XC")),

some instances fixed where "coordinate" was spelled "coordinates".  May have no effect.

### DIFF
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -76,7 +76,7 @@ horizontal_coordinates_spherical = OrderedDict(
         units="degrees_east", coordinate="YG XG")),
     YG=dict(dims=["j_g", "i_g"], attrs=dict(
         standard_name="latitude_at_f_location", long_name="latitude",
-        units="degrees_north", coordinates="YG XG"))
+        units="degrees_north", coordinate="YG XG"))
 )
 
 horizontal_coordinates_llc = horizontal_coordinates_spherical.copy()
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks so much for this Ian!

Can you think of a way to test that this feature is working properly? For example, perhaps some check in this test?

https://github.com/MITgcm/xmitgcm/blob/7c5305cc81dee6a21e0b035e8682d03c113c10de/xmitgcm/test/test_mds_store.py#L93-L98

rabernat:
Ian, can you help us understand what impact this PR has? I agree there is a typo, but I'm struggling to understand the consequences of that typo.

If this `coordinates` / `coordinates` entry actually has no effect, we should just remove it.

ifenty:
@rabernat Consequences are probably zero with respect to any calculations because when the dataset is saved to netcdf, xarray creates a 'coordinate' attribute for data variables: a 'space separated list of auxillary coordinates' as per CF conventions. It seems that xmitgcm is specifing its own coordinate list (e.g.,  ['XC YC'] or ['XG YG']) which xarray **might** use instead of making its own list.  Because of the typo, the saved netcdf dataset **could** have a variables with both a 'coordinates' attributes (misspelled coordiante added by xmitgcm) and a 'coordinate' attribute added by xarray. I believe this  'coordinates' attribute is used in some plotting software, such as panoply.  Seems like it's ok to keep the code in 'xmitgcm'.

rabernat:
Ok understood. I'd prefer to have some sort of test for this, but since it's a very minor change, we can merge this as is.

Thanks!

Issue #234, opened 2020-11-06
Error when trying to load only 'U' or 'V' from an llc dataset


fraserwg:
I've been trying to run the following code to extract 'U' or 'V' from the llc4320 dataset.

```python
import xmitgcm.llcreader as llcreader

model = llcreader.ECCOPortalLLC4320Model()
ds_vel = model.get_dataset(varnames=['U'], type='latlon', iter_start=10368, iter_stop=10369)
```

If `varnames` is set to either `['U']` or [`'V'`] an error message is received. If it is set to `['U', 'V']` the function runs as expected. This isn't a problem if loading other variables. on their own such as `'Eta'` or `'W'`.

The error message is below:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-9a90c8e0ac18> in <module>
----> 1 ds_vel = model.get_dataset(varnames=['U'], type='latlon', iter_start=10368, iter_stop=10369)

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/xmitgcm-0.4.1+20.g7c5305c-py3.8.egg/xmitgcm/llcreader/llcmodel.py in get_dataset(self, varnames, iter_start, iter_stop, iter_step, k_levels, k_chunksize, type, read_grid, grid_vars_to_coords)
    727 
    728         transformer = data_transformers[type]
--> 729         data = transformer(data_facets, _VAR_METADATA)
    730 
    731         # separate horizontal and vertical grid variables

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/xmitgcm-0.4.1+20.g7c5305c-py3.8.egg/xmitgcm/llcreader/llcmodel.py in _all_facets_to_latlon(data_facets, meta)
    354             mate = meta[vname]['attrs']['mate']
    355             vector_pairs.append((vname, mate))
--> 356             vnames.remove(mate)
    357         except KeyError:
    358             pass

ValueError: list.remove(x): x not in list
```

There is also no problem if `type` is set to `'faces'`

rabernat:
This is an interesting issue.

The challenge here is that U and V are actually tightly coupled together in the LLC4320 internal data storage format. When transforming to the `latlon` grid, we mix together U and V. So it's not possible to get one without the other.

I think the best path forward would be to add a quick check in `get_dataset` to ensure that all the metric vector pairs are present and raise an error if not. Would that satisfy you?

fraserwg:
[...]

PR #237, opened 2020-11-12
feat: IMPLEMENTS CHECKING FOR VARIABLE MATES

fraserwg:
On a latlon LLC grid some variables have mates. This commit will produce a meaningful error if attempting to use `get_dataset` or `faces_dataset_to_latlon` to load a variable without its mate.

See Github issue #234

### DIFF
--- a/xmitgcm/llcreader/llcmodel.py
+++ b/xmitgcm/llcreader/llcmodel.py
@@ -298,13 +298,21 @@ def faces_dataset_to_latlon(ds, metric_vector_pairs=[('dxC', 'dyC'), ('dyG', 'dx
     vector_pairs = []
     scalars = []
     vnames = list(ds.reset_coords().variables)
+
     for vname in vnames:
         try:
             mate = ds[vname].attrs['mate']
-            vector_pairs.append((vname, mate))
-            vnames.remove(mate)
         except KeyError:
-            pass
+            mate = None
+
+        # Raises an exception if the mate of a variable in vnames is missing.
+        if mate is not None:
+            vector_pairs.append((vname, mate))
+            try:
+                vnames.remove(mate)
+            except ValueError:
+                msg = 'If {} in varnames, {} must also be in varnames'.format(vname, mate)
+                raise ValueError(msg)
 
     all_vector_components = [inner for outer in (vector_pairs + metric_vector_pairs)
                              for inner in outer]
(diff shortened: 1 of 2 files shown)


xmitgcm/llcreader/llcmodel.py
```diff
@@ -349,13 +357,21 @@ def _all_facets_to_latlon(data_facets, meta):
...
+            mate = None
+
+        # Raises an exception if the mate of a variable in vnames is missing.
+        if mate is not None:
+            vector_pairs.append((vname, mate))
+            try:
+                vnames.remove(mate)
+            except ValueError:
+                msg = 'If {} in varnames, {} must also be in varnames'.format(vname, mate)
+                raise ValueError(msg)
```
rabernat, line 374: This PR makes me realize we have some duplicative code paths in here. I'm trying to understand why we need to implement this in both `_all_facets_to_latlon` and `faces_dataset_to_latlon`. I think that `faces_dataset_to_latlon` should probably be refactored to call `_all_facets_to_latlon`. However, that is not your problem.

rabernat (changes requested):
This looks great! Thank you so much.

We will just need a test for this scenario. It would go in here: https://github.com/MITgcm/xmitgcm/blob/7c5305cc81dee6a21e0b035e8682d03c113c10de/xmitgcm/test/test_llcreader.py#L76

Something like
```python
def test_vector_mate_error(local_llc90_store):
    store = local_llc90_store
    model = llcreader.LLC90Model(store)
    with pyest.raises(ValueError, match=r".* must also be .*"):
        ds_latlon = model.get_dataset(type='latlon', varnames=['U'], iter_start=0, iter_stop=9, iter_step=8)
```

fraserwg:
I've added the test function but am now getting test fails on the Xarray-master build, details below:

```python
=========================== short test summary info ============================
257
FAILED xmitgcm/test/test_llcreader.py::test_ecco_portal_load[2160] - ValueErr...
258
FAILED xmitgcm/test/test_llcreader.py::test_ecco_portal_load[4320] - ValueErr...
259
====== 2 failed, 276 passed, 3 skipped, 124 warnings in 106.82s (0:01:46) ======
```

The tests weren't failing for the previous commit and I haven't changed the tests that are now failing. Also the other builds are passing (as are the tests on my local machine) -- could this be a problem with a new commit on the Xarray-master branch? Any ideas for where to start with debugging would be appreciated, I'm relatively new to the world of CI and pytest!

fraserwg:
Have just added a new commit with the test removed again. Looks like everything's passed suggesting the problem is indeed with my test function.

fraserwg:
Ok, all the problems have been sorted, it was just a missing pair of brackets :)

rabernat:
Fantastic! Thanks so much. I appreciate your patience to sort out the CI failures.

Issue #235, opened 2020-11-06
Meaningful errors for non-existent iterations in llcreader


fraserwg:
If setting iter_start and iter_stop to a range where no time steps exist it would be useful to get an error message saying this is a problem. For instance the below code will try and load iteration 1 from the llc4320 dataset:

```python
import xmitgcm.llcreader as llcreader

model = llcreader.ECCOPortalLLC4320Model()

ds_vel = model.get_dataset(varnames=['Eta'], type='latlon', k_levels=[1], iter_start=1, iter_stop=2)

ds_vel.compute()
```

Upon calling the compute function the following error message is shown:

```python
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-21-d54b1fa42a40> in <module>
----> 3 ds_vel.compute()

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/xarray/core/dataset.py in compute(self, **kwargs)
    817         """
    818         new = self.copy(deep=False)
--> 819         return new.load(**kwargs)
    820 
    821     def _persist_inplace(self, **kwargs) -> "Dataset":

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/xarray/core/dataset.py in load(self, **kwargs)
    652 
    653             # evaluate all the dask arrays simultaneously
--> 654             evaluated_data = da.compute(*lazy_data.values(), **kwargs)
    655 
    656             for k, data in zip(lazy_data, evaluated_data):

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/dask/base.py in compute(*args, **kwargs)
    450         postcomputes.append(x.__dask_postcompute__())
    451 
--> 452     results = schedule(dsk, keys, **kwargs)
    453     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    454 

~/anaconda3/envs/xmitgcm_env/lib/python3.8/site-packages/distributed/client.py in get(self, dsk, keys, restrictions, loose_restrictions, resources, sync, asynchronous, direct, retries, priority, fifo_timeout, actors, **kwargs)
   2723                     should_rejoin = False
   2724             try:
-> 2725                 results = self.gather(packed, asynchronous=asynchronous, direct=direct)
   2726             finally:
   2727                 for f in futures.values():

[...]

PR #240, opened 2020-11-12
feat: CHECKS ITERATION PARAMETERS WHEN GETTING LLCMODEL DATASETS

fraserwg:
When loading datasets using the llcreader the user is able to specify the iteration start, stop and step. These can also be specified as attributes in the `BaseLLCModel`. This commit introduces a check that compares the `iter_start` and `iter_step` fed into `get_dataset` with the `BaseLLCModel` attributes. If they are found to be incompatible a warning is triggered.

A warning over an Error was chosen as subsuquent code may still execute successfully even if there is a mismatch between the user specified and `BaseLLCModel` derived values for `iter_start` and `iter_stop`.


fraserwg:
This implements my request from issue #235

rabernat:
Thanks a lot for this! We really appreciate your willingness to dive into xmitgcm.

We strive for high test coverage for this package. Therefore, you will need to add a test for this new feature. You will probably want to use the pytest warnings mechanism: https://docs.pytest.org/en/3.0.5/recwarn.html.

rabernat:
It looks like there is now a merge conflict here, since we recently merged #231. Perhaps @timothyas can help you sort it out?

timothyas:
No problem, it seems like the merge conflicts are all gone - thanks @fraserwg! I think this looks good to me, I just have one minor request to add to the tests, but then I think it's good to go.

timothyas (changes requested):
Could you add a test which checks for this warning while calling `get_dataset`? For example at the bottom of this test:
https://github.com/MITgcm/xmitgcm/blob/04ec9fb7a819181545dcb455f8553bb5695bbcf6/xmitgcm/test/test_llcreader.py#L127
just another `with pytest.warns` check as you did for the others

xmitgcm/test/test_llcreader.py
```diff
@@ -165,6 +166,25 @@ def test_ecco_portal_latlon(ecco_portal_model):
             assert len(ds_ll[fld].data.chunks)==1
             assert (len(ds_ll[fld]),)==ds_ll[fld].data.chunks[0]
 
+
+def test_iteration_start_check(ecco_portal_model):
+    with pytest.warns(RuntimeWarning) as record:
+        ecco_portal_model._check_iter_start(ecco_portal_model.iter_start - 1)  # Should give runtime warning
+        ecco_portal_model._check_iter_start(ecco_portal_model.iter_start)  # Should give no warning
+        ecco_portal_model._check_iter_start(ecco_portal_model.iter_start + ecco_portal_model.iter_step) # Should give no warning
```
rabernat, line 174: This is not quite what I had in mind for a test. You should not call private functions (e.g. `._check_iter_start`) from the test suite. Instead, you should call the public API (e.g. `get_dataset`) as suggested by @timothyas.

rabernat (changes requested):
Thanks for your work on this @fraserwg and for your patience as we review your PR.

xmitgcm/test/test_llcreader.py
fraserwg: Sure, I've rewritten the tests to call `get_dataset` but the ECCO Data Portal is currently down. I'll push the changes when I've verified they work locally 👍

rabernat:
I think the test failure is a fluke, so I'm going to merge this now. Thanks for another great contribution!

PR #98, opened 2018-09-18
Cube Sphere support

raphaeldussin:
@rabernat @brian-rose I wrote the low level function to read chunks for the Cube Sphere for issue #91. It works with limited capabilities and needs more testing.
Next steps:

- [x] add the CS configuration in the tests
- [x] add the CS geometry in the high level functions

PR #246, opened 2021-01-06
full size xg and yg

AaronDavidSchneider:
This PR picks up on PR #195 and is currently WIP.
I plan to keep this PR minimal and not pick up the ability to reshape a dataset in tiles of a specified shape.
This PR only extends xg and yg to their full size.


AaronDavidSchneider:
I will delay this PR until PR #98 is merged. 
It makes no sense to do the work twice 😀

AaronDavidSchneider:
Hi @rabernat,  

I think I found a solution. In the current version of this PR we are extending extending i_g and j_g to nx + 1.
I think my implementation is straight forward and quite easy.

I tested it with a cs grid. But it should also work with llc.

Cheers
Aaron

xmitgcm/utils.py
```diff
@@ -1423,7 +1443,10 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
                     if grid_metadata['face_facets'][face] == kfacet:
                         # get offset of face from facet
                         offset = file_metadata['face_offsets'][face]
-                        nx = file_metadata['nx']
+                        if field in outerx_vars or field in outerxy_vars or field in outery_vars:
```
raphaeldussin, line 1446: or a more concise way:
```python
if field in outerx_vars + outerxy_vars + outery_vars:
```

xmitgcm/utils.py
```diff
@@ -1462,9 +1486,9 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
                           coords={'i': (['i'], np.arange(file_metadata['nx'])),
                                   'j': (['j'], np.arange(file_metadata['nx'])),
                                   'i_g': (['i_g'],
-                                          np.arange(file_metadata['nx'])),
+                                          np.arange(nxouter)),
                                   'j_g': (['j_g'],
-                                          np.arange(file_metadata['nx'])),
+                                          np.arange(nxouter)),
```
raphaeldussin, line 1491: this is only true because nx = ny in CS and LLC **faces**

raphaeldussin:
I think this PR solves the issues of my previous implementation of this utility function

xmitgcm/utils.py
AaronDavidSchneider: that makes sense. I changed it for the else case (not cs and not llc).
It seems that it was also wrong before

AaronDavidSchneider (commented):
Thanks @raphaeldussin for your feedback! I changed it to match your feedback.

AaronDavidSchneider:
Hi @rabernat, @raphaeldussin, just a quick reminder that my work on this PR is finished.

AaronDavidSchneider:
Hi, just another casual reminder @rabernat and @raphaeldussin (sorry for bothering).

rabernat:
Thanks for the reminder @AaronDavidSchneider, and sorry for taking so long to review! We really appreciate your contribution. I will do my best to have a look this week.

xmitgcm/utils.py
```diff
@@ -1413,7 +1425,15 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
 
             for field in file_metadata['fldList']:
                 # symetrize
-                tmp = rawfields[field][:, :, :-1, :-1].squeeze()
+                if field in outerx_vars:
+                    tmp = rawfields[field][:, :, :, :-1].squeeze()
```
rabernat, line 1429:
```suggestion
                    tmp = rawfields[field][..., :-1].squeeze()
```

xmitgcm/utils.py
```diff
@@ -1413,7 +1425,15 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
...
+                elif field in outery_vars:
+                    tmp = rawfields[field][:, :, :-1, :].squeeze()
```
rabernat, line 1431:
```suggestion
                    tmp = rawfields[field][..., :-1, :].squeeze()
```

xmitgcm/utils.py
```diff
@@ -1413,7 +1425,15 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
...
+                elif field in outerxy_vars:
+                    tmp = rawfields[field][:, :, :, :].squeeze()
```
rabernat, line 1433:
```suggestion
                    tmp = rawfields[field].squeeze()
```

xmitgcm/utils.py
```diff
@@ -1413,7 +1425,15 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
...
+                else:
+                    tmp = rawfields[field][:, :, :-1, :-1].squeeze()
```
rabernat, line 1435:
```suggestion
                    tmp = rawfields[field][..., :-1, :-1].squeeze()
```

rabernat (changes requested):
This is really great!

My one central concern is that we don't have a test. @raphaeldussin - do any of the test cases contain the needed input files to test this code?

AaronDavidSchneider:
Thanks for your comments @rabernat!

AaronDavidSchneider:
Regarding the tests:
It should be feasible to check if the XC and YC values are in the middle of the XG and YG values of outer=True. 

What do you think?

rabernat:
So I looked it up, and it seems that we already have a good test for the `get_grid_from_input`

https://github.com/MITgcm/xmitgcm/blob/4a9c748c2e7cf140d85efee8851cac9d7e7c45e0/xmitgcm/test/test_utils.py#L961

To this test you should add a section that calls the function with `outer=True`.

> It should be feasible to check if the XC and YC values are in the middle of the XG and YG values of outer=True.

No, I would not recommend that.  The data are coming from files hosted in the test data repo, so I would not make assumptions about the values. You just want to check that the arrays are the right shape.

Do you know how to run the test suite? Just run `pytest` from the root directory.

raphaeldussin:
@rabernat we should have the CS grid on figshare, see https://github.com/MITgcm/xmitgcm/blob/4a9c748c2e7cf140d85efee8851cac9d7e7c45e0/xmitgcm/test/test_xmitgcm_common.py#L137

let me know if there is an issue with it

AaronDavidSchneider:
I added some tests that fail. However, I don't know why. I have the bad feeling that outer=True only works for the cs geometry.
I think I need help with fixing the bug, since I am not familiar with aste or llc :D 
Thank you for your time!

rabernat:
> I added some tests that fail. However, I don't know why. I have the bad feeling that outer=True only works for the cs geometry.

This is why we have tests: to find cases we did not consider before users do!

I am hoping @raphaeldussin can help you fix the bug. He knows this part of the code base the best. For reference, here is the test failure:

<details>

```
        ds_outer = get_grid_from_input(dirname + '/' + expected['gridfile'],
                                 geometry=expected['geometry'],
                                 dtype=np.dtype('d'), endian='>',
                                 use_dask=usedask,
                                 extra_metadata=md,
>                                outer=True)

xmitgcm/test/test_utils.py:978: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
xmitgcm/utils.py:1452: in get_grid_from_input
    padded = _pad_array(tmp, file_metadata, face=face)
xmitgcm/utils.py:1229: in _pad_array
    (pad_before, data), axis=0)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = ((array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
  ...       ],
       [-38.        , -37.93939671, -37.8460971 , ...,  51.8460971 ,
         51.93939671,  52.        ]])),)
kwargs = {'axis': 0}
relevant_args = (array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
   ....        ],
       [-38.        , -37.93939671, -37.8460971 , ...,  51.8460971 ,
         51.93939671,  52.        ]]))

>   ???
E   ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 270 and the array at index 1 has size 271
```

</details>

AaronDavidSchneider:
Hi @raphaeldussin,
could you give me some feedback on how we can fix this PR for llc geometry?
Alternatively, we could raise a notimplementederror if the outer argument is used with the llc geometry.
Thanks!

raphaeldussin:
https://github.com/AaronDavidSchneider/xmitgcm/pull/1 should solve the problem.

AaronDavidSchneider:
Thanks @raphaeldussin for your feedback and work on this PR.
It should work now.

xmitgcm/test/test_utils.py
```diff
@@ -963,26 +963,50 @@ def test_get_grid_from_input(all_grid_datadirs, usedask):
...
     dirname, expected = all_grid_datadirs
     md = get_extra_metadata(domain=expected['domain'], nx=expected['nx'])
+
     ds = get_grid_from_input(dirname + '/' + expected['gridfile'],
                              geometry=expected['geometry'],
                              dtype=np.dtype('d'), endian='>',
                              use_dask=usedask,
                              extra_metadata=md)
+
+    ds_outer = get_grid_from_input(dirname + '/' + expected['gridfile'],
```
raphaeldussin, line 973: we could make the test more compact by using a pytest.mark.parameterize on outer

xmitgcm/test/test_utils.py
```diff
@@ -963,26 +963,50 @@ def test_get_grid_from_input(all_grid_datadirs, usedask):
...
+        if var in outery_vars or var in outerxy_vars:
+            expected_shape_outer[-2] = expected_shape_outer[-2] + 1
+
         assert type(ds[var]) == xarray.core.dataarray.DataArray
         assert ds[var].values.shape == expected['shape']
+        assert type(ds_outer[var]) == xarray.core.dataarray.DataArray
+        assert ds_outer[var].values.shape == tuple(expected_shape_outer)
 
     # check we don't leave points behind
     if expected['geometry'] == 'llc':
```
raphaeldussin, line 1012: we would also want to test this with cs geometry

xmitgcm/utils.py
```diff
@@ -1347,6 +1349,15 @@ def get_grid_from_input(gridfile, nx=None, ny=None, geometry='llc',
                                 'XG', 'YG', 'DXV', 'DYU', 'RAZ',
                                 'DXC', 'DYC', 'RAW', 'RAS', 'DXG', 'DYG']
 
+    if outer:
```
raphaeldussin, line 1352: suggest using outer_vars = ['DXC',...] if outer else []

raphaeldussin (changes requested):

raphaeldussin:
getting very close to the finish line @AaronDavidSchneider

xmitgcm/test/test_utils.py
AaronDavidSchneider: I think I don't understand what you mean here.

xmitgcm/test/test_utils.py
AaronDavidSchneider: Good Idea. Did that.

xmitgcm/utils.py
AaronDavidSchneider: Good idea.

AaronDavidSchneider:
Thanks for your feedback @raphaeldussin. I tried to address the proposed changes. I still have one open question (see above).

raphaeldussin (changes requested):

AaronDavidSchneider:
So I just had a look into the llc checks in `test_get_grid_from_input` and tried to come up with a simple test that would do the same for the cs geometry.
However, the cs geometry is slightly different and I would need more knowledge about the raw data. E.g.
```py
xc1 = read_raw_data(grid1, dtype=np.dtype('>d'), shape=(ny1, nx),
                            partial_read=True)
```
does not work.

I would suggest, that we merge this PR first (once its ready) and work on cs grid checks afterwards, since these checks are not really related to this PR anyway.


I would be happy to open up a new PR to work on these checks. However, I would need your help @raphaeldussin for the `read_raw_data` part.

raphaeldussin:
I think this test on the values is important so we preserve the answers in any future code changes.
inferring/reading the raw data can do this way: 

`du -b file.bin` and divide by 8 (number of bytes in double precision, which is the format used for grids), then nx+1 = 33 twice
since you have square faces of cs32 and you get 18 which is your number of variables. Then

read all variables:
```python
raw = read_raw_data('grid_cs32.face001.bin', np.dtype('>d'), (18,33,33))
```

read only first one:
```python
raw = read_raw_data('grid_cs32.face001.bin', np.dtype('>d'), (33,33), partial_read=True)
```

read second:
```python
raw = read_raw_data('grid_cs32.face001.bin', np.dtype('>d'), (33,33), partial_read=True, offset=(33*33*8))
```

from there you should be able to add a section to the test for CS geometry.

AaronDavidSchneider:
Thanks for the explanation @raphaeldussin! I took the llc check and adapted it for the CS grid. What do you think?

raphaeldussin (approved):

raphaeldussin:
LGTM! @rabernat are you good with this as well?

AaronDavidSchneider:
Hi @rabernat, could you have a final look on this PR?
Thanks!

AaronDavidSchneider:
Hi @rabernat, could you have a final look on this PR?
Thanks!

AaronDavidSchneider:
Just another kind reminder @rabernat.

AaronDavidSchneider:
Hi @rabernat,
is there any possibility that this could be merged soon?

xmitgcm/test/test_utils.py
```diff
@@ -1033,13 +1047,74 @@ def test_get_grid_from_input(all_grid_datadirs, usedask):
...
+        grid = expected['gridfile']
+        grid1 = dirname + '/' + grid.replace('<NFACET>', '001')
+        grid2 = dirname + '/' + grid.replace('<NFACET>', '002')
+        grid3 = dirname + '/' + grid.replace('<NFACET>', '003')
+        grid4 = dirname + '/' + grid.replace('<NFACET>', '004')
+        grid5 = dirname + '/' + grid.replace('<NFACET>', '005')
+        grid6 = dirname + '/' + grid.replace('<NFACET>', '006')
+
+
+        xc1 = read_raw_data(grid1, dtype=np.dtype('>d'), shape=(nx, nx),
```
rabernat, line 1079: Is this dtype (`'>d'`) always the same for every mitgcm run everywhere? If not, we will need a parameter to configure it.

rabernat (commented):
Sorry for being extremely slow here. Just one question, then LGTM.

xmitgcm/test/test_utils.py
AaronDavidSchneider: Oh well that's a good question! I just used what @raphaeldussin suggested above. Maybe @raphaeldussin knows better?

raphaeldussin: It would be easy to add extra keys (`endianness`, `precision`) to _grids here https://github.com/MITgcm/xmitgcm/blob/c793ecceba5393aac9be0bbdf91a53a3175ecccc/xmitgcm/test/test_xmitgcm_common.py#L175 and pass them down to the test.

AFAIK all the grid files are written in double precision and big-endian, I'm not sure MITgcm can read any other format.

rabernat: > AFAIK all the grid files are written in double precision and big-endian, I'm not sure MITgcm can read any other format.

If this is true, then we can save ourselves the trouble and just merge this as is. Is there any evidence in the MITgcm source code?

AaronDavidSchneider: I think MITgcm is per default big endian and double precision.
But I think you could in principle change it.

raphaeldussin: let's put in the extra keys, I doubt we'll ever use other values but we'll get it covered. Did not find hard restrictions in MITgcm so far.

AaronDavidSchneider: Sounds good to me. Will do so.

AaronDavidSchneider:
Added precision and endianness to grid common.
Thanks a lot @rabernat and @raphaeldussin!

rabernat (approved):
Fantastic!

Reading the code, I realized I don't like how we have separate parameters for `dtype` and `endian`. We can handle endianness as part of dtype, so this is unnecessary. However, that predates your PR, so it doesn't have to be fixed here.

PR #247, opened 2021-01-07
add a standard set of ADJ variables

mjlosch:
for my ecco-like adjoint simulations, I would like to be able to read the hard coded standard adjoint output (per adjDumpFreq) with xmitgcm. Appending the dict `packages_state_variables` let's me do so.

Also there's s small fix to one variable name and unit.

### DIFF
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -602,8 +602,151 @@ package_state_variables = {
         units='W m-2')),
     'QSWtave': dict(dims=['j', 'i'], attrs=dict(
         standard_name="surface_net_upward_shortwave_flux",
-        long_name='SEAICE upward freshwater flux (+=up)',
-        units='kg m-2 s-1'))
+        long_name='SEAICE upward shortwave heat flux (+=up)',
+        units='W m-2')),
+    # pkg/autodiff variables
+    'ADJtaux': dict(dims=['j', 'i_g'], attrs=dict(
+        standard_name="ADJtaux",
+        long_name='dJ/dtaux: Sensitivity to zonal surface stress',
+        mate='ADJtauy',
+        units='dJ/(N m-2)')),
+    'ADJtauy': dict(dims=['j_g', 'i'], attrs=dict(
+        standard_name="ADJtaux",
+        long_name='dJ/dtauy: Sensitivity to meridional surface stress',
+        mate='ADJtaux',
+        units='dJ/(N m-2)')),
+    'ADJqnet': dict(dims=['j', 'i'], attrs=dict(
+        standard_name="ADJqnet",
+        long_name='dJ/dqnet: Sensitivity to net upward heat flux',
+        units='dJ/(W m-2)')),
+    'ADJempr': dict(dims=['j', 'i'], attrs=dict(
+        standard_name="ADJempr",
+        long_name='dJ/dempr: Sensitivity to upward freshwater flux',
+        units='dJ/(kg m-2 s-1)')),
+    'ADJqsw': dict(dims=['j', 'i'], attrs=dict(
+        standard_name="ADJqsw",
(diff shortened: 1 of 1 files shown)


timothyas:
Thanks @mjlosch! This looks good to me. 

As an aside, adjoint variables can also be printed through the diagnostics package of the MITgcm. This means that they get written to `available_diagnostics.log` so that xmitgcm will recognize them. I've been doing this for any new control variable I've added and it's fairly smooth. A future todo that I haven't gotten to yet over at MITgcm is to add all of these fairly standard adjoint variables to that list. Not hard, just tedious.

That's just some rambling though, this is still very useful because I'm not sure anyone knows about or uses the adjoint+diagnostics capabilities.

--- 

PS I'm hoping to get to your PR reviews and my open PRs over there next week

mjlosch:
Thanks @timothyas, I saw that something was already done for the 5 basic state variables, but it was not clear to me that this is universally possible. That's of course much better than what I have attempted here. It would be very good to have these "standard" diagnostics included in the diagnostics scheme (currently I am filling up my disks with all this unwanted output). If you already have started doing this somewhere, I'd be happy to help shape this into a contribution to MITgcm, maybe just for the variables I have added here?

timothyas:
That sounds great. I can start a draft PR next week when I get to the other MITgcm PRs that I need to tend to. I'd honestly like to overhaul ad_dummy_in_stepping while we're at it, because it's getting a little ridiculous, and the way it's written makes adding control variables more laborious than necessary.

timothyas:
But, in relation to this PR: LGTM

xmitgcm/variables.py
```diff
@@ -602,8 +602,151 @@
...
+        standard_name="ADJggl90tke",
+        long_name='dJ/dggl90tke: Sensitivity to TKE',
+        units='dJ/(m2 s-2)')),
+    'ADJdiffkr': dict(dims=['k', 'j', 'i'], attrs=dict(
+        standard_name="ADJdiffkr",
+        long_name='dJ/ddiffkr: Sensitivity to vertical diffusivity',
+        units='dJ/(m2 s-1)')),
+    'ADJkapgm': dict(dims=['k', 'j', 'i'], attrs=dict(
+        standard_name="ADJkapgm",
+        long_name='dJ/dkapgm: Sensitivity to meridional surface stress',
```
timothyas, line 640: Should be e.g. `sensitivity to GM Intensity`

xmitgcm/variables.py
```diff
@@ -602,8 +602,151 @@
...
+        units='dJ/(m2 s-1)')),
+    'ADJkapredi': dict(dims=['k', 'j', 'i'], attrs=dict(
+        standard_name="ADJkapredi",
+        long_name='dJ/dkapredi: Sensitivity to meridional surface stress',
```
timothyas, line 644: e.g. `sensitivity to Redi coefficient`

rabernat:
Great! Thanks a lot @mjlosch!

I see no problem with adding as much metadata as we can. My only concern is test coverage. We don't have an example that includes these fields, so we won't really know if there is a problem with them.

Let's think about how to solve that problem in the long run. In the meantime, LGTM!

rabernat:
Expanding a bit...it would be great if there were a central source of truth for MITgcm variable metadata that both xmitgcm and MITgcm itself could pull from. Perhaps a namelist file in the MITgcm source code, or a standalone database in yaml/json?

rabernat:
Ug, sorry, it looks like I merged prematurely. I saw @timothyas "LGTM" and just clicked merge.

I picked this up in #249. Any more typos we should fix over there?

timothyas:
Sorry, that was my bad! I had just skimmed it but noticed the typos while working on the MITgcm PR mentioned above. I added some fixes over there. 

+1 for a variable metadata truth file :)

PR #248, opened 2021-01-10
Update xarray master ci

timothyas:
Simply update CI workflow to use python 3.7, which xarray master now requires.

Should we drop python 2.7? phase out 3.6? I'd be happy to do that here, but not sure what it entails other than:

- [x] drop the CI workflows
- [x] remove from README etc

It is maybe not critical to remove stuff like:
```
if (sys.version_info < (3,0):
    do something
```
But this would clean things up. I'm curious what you think.


rabernat:
Thanks so much for this fix @timothyas!

> Should we drop python 2.7? phase out 3.6? 

👍  to both!

> not sure what it entails

Maybe we would have to update setup.py in some places.

rabernat:
Generally speaking, there is no reason for us to be more conservative with our dependencies than Xarray. We can just follow what Xarray does in terms of python version support.

timothyas:
No problem!

> Generally speaking, there is no reason for us to be more conservative with our dependencies than Xarray. We can just follow what Xarray does in terms of python version support.

That definitely makes sense. I'll get to this as soon as I can :)

timothyas:
Ok I just finished:
- updating the URL and dependencies in setup.py: specified python>=3.7 in setup, and xarray >=0.11.0 since this is when they introduced python 3.7
- updated versions of python, xarray, dask in readme and installation documentation
- drop CI for 2.7, 3.6, add CI for 3.9

I think it's good to go for whenever you get the chance to check it out.

rabernat:
LGTM!

PR #249, opened 2021-01-12
fix variable metadata typos

rabernat:
Picks up where #247 left off.

### DIFF
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -637,7 +637,7 @@ package_state_variables = {
         units='dJ/(m2 s-1)')),
     'ADJkapgm': dict(dims=['k', 'j', 'i'], attrs=dict(
         standard_name="ADJkapgm",
-        long_name='dJ/dkapgm: Sensitivity to meridional surface stress',
+        long_name='dJ/dkapgm: Sensitivity to GM Intensity',
         units='dJ/(m2 s-1)')),
     'ADJkapredi': dict(dims=['k', 'j', 'i'], attrs=dict(
         standard_name="ADJkapredi",
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -625,8 +625,8 @@ package_state_variables = {
         units='dJ/(kg m-2 s-1)')),
     'ADJqsw': dict(dims=['j', 'i'], attrs=dict(
         standard_name="ADJqsw",
-        long_name='dJ/dqsw: Sensitivity to net upward heat flux',
-        units='dJ/(N m-2)')),
+        long_name='dJ/dqsw: Sensitivity to net shortwave radiation',
+        units='dJ/(W m-2)')),
     'ADJggl90tke': dict(dims=['k', 'j', 'i'], attrs=dict(
         standard_name="ADJggl90tke",
         long_name='dJ/dggl90tke: Sensitivity to TKE',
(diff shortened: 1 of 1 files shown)


timothyas:
Thanks for taking care of this!

rabernat:
When you make a 🤦  the best thing to do is fix it!

xmitgcm/variables.py
```diff
@@ -712,19 +712,19 @@
     'ADJprecip': dict(dims=['j', 'i'], attrs=dict(
         standard_name="ADJprecip",
         long_name='dJ/dprecip: Sensitivity to precipitation flux',
-        units='dJ/(kg m-2 s-1)')),
+        units='dJ/(m s-1)')),
     'ADJrunoff': dict(dims=['j', 'i'], attrs=dict(
         standard_name="ADJrunoff",
         long_name='dJ/drunoff: Sensitivity to runoff',
-        units='dJ/(kg m-2 s-1)')),
+        units='dJ/(m s-1)')),
```
timothyas, line 719: @mjlosch can you confirm the ADJ precip and runoff units? I made these changes based on what they are in the exf package.

timothyas:
I just went through and made some changes directly. I think it's good once @mjlosch confirms a couple of variables as noted

xmitgcm/variables.py
mjlosch: @timothyas you are right, that was another oversight at my end.

xmitgcm/variables.py
```diff
@@ -712,19 +712,19 @@
...
     'ADJclimsst': dict(dims=['j', 'i'], attrs=dict(
         standard_name="ADJclimsst",
         long_name='dJ/dclimsst: Sensitivity to restoring surface temperature',
         units='dJ/K')),
     'ADJclimsss': dict(dims=['j', 'i'], attrs=dict(
         standard_name="ADJclimsss",
         long_name='dJ/dclimsss: Sensitivity to restoring surface salinity',
-        units='dJ/pss)')),
+        units='dJ/psu)')),
```
mjlosch, line 727: Here I disagree, although my solution is not optimal either. Salinity has no units. In particular "psu" (practical salinity unit) is something that should not be used. pss = practical salinity scale is a compromise but also not very good, see e.g. https://www.myroms.org/forum/viewtopic.php?t=294 
If we want to use absolute salinity, the "unit" is g/kg, which again is 1/1000 with no units. In https://github.com/MITgcm/MITgcm/blob/master/model/inc/DYNVARS.h we have "ppt" (parts per thousand), which is again not a unit but just a ratio as it should be. To be consistent, we can change this to dJ/ppt if you like.

mjlosch (changes requested):
Except for "psu" I am fine with the changes and apologise for my sloppiness.

xmitgcm/variables.py
timothyas: Thanks for another look at all of this @mjlosch! I don't have a strong opinion about what the "units" (or not!) of salinity are labeled as. I put `psu` because at the end of the day it is what the MITgcm diagnostics assigns to the variable `SALT` (and anything involving it like advection of salt, etc), at least for the setups I am using. I see what you're saying about `DYNVARS.h`. My vote is for consistency, and so if we go for e.g. `ppt` here, I would be in favor of opening this discussion in the MITgcm as well to change it there too.

Unless I'm missing something here... Let me know what you think.

mjlosch: We should probably go for g/kg, and then open an issue on MITgcm, where the comment need to be changed in DYNVARS.h and the units in diagnostics_main_init.F
@rabernat please decide on the "unit"/non-unit. We can always change it, once the MITgcm code is settled.

mjlosch: @timothyas @rabernat 
Yesterday, I talked to @jm-c, @christophernhill, @jrscott, @jahn and we agreed to replace the "units" of salinity by "g/kg" in the MITgcm code, so I suggest to do this here as well anticipating this change ("in vorauseilendem Gehorsam"). I can do that, if you agree.

timothyas: That sounds good to me, I'm in favor of g/kg. If @rabernat agrees, we'll want to change it for every salinity instance in variables.py. Thanks @mjlosch

mjlosch (approved):

timothyas:
I'm having a hard time trusting my eyes these days, but this looks good to me 👍

rabernat:
Thanks everybody. I really appreciate the work.

PR #250, opened 2021-01-12
fixed mistake with naming of Zl/Zu and k_l/k_u.  

ifenty:
Now l='lower', u='upper' in the descriptions of four variables: Zl, Zu, k_l, k_u.

@rabernat


rabernat:
Excellent! Thanks @ifenty.

The CI failure on the xarray-master env is unrelated:

```
Pip subprocess error:
  ERROR: Package 'xarray' requires a different Python: 3.6.12 not in '>=3.7'
```

We will need to fix that somehow eventually... Package maintenance never ends. 🙃

Issue #254, opened 2021-01-15
Trouble interacting with coordinate variables on Pleiades


dhruvbalwada:
When I try to access the values of the coordinate variables on Pleiades, using the llcreader I get an error.

To reproduce:

```
model = llcreader.PleiadesLLC4320Model() # same error with 2160 too.
ds = model.get_dataset(varnames=['Salt','Theta'], type='latlon')
ds.XC.sel(i=5000).plot() # can try .values and get same error too.
```

The error is: 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-3c50e58ba8dd> in <module>
----> 1 ds.Depth.sel(i=5000).plot()

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/xarray/plot/plot.py in __call__(self, **kwargs)
    444 
    445     def __call__(self, **kwargs):
--> 446         return plot(self._da, **kwargs)
    447 
    448     # we can't use functools.wraps here since that also modifies the name / qualname

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/xarray/plot/plot.py in plot(darray, row, col, col_wrap, ax, hue, rtol, subplot_kws, **kwargs)
    161 
    162     """
--> 163     darray = darray.squeeze().compute()
    164 
    165     plot_dims = set(darray.dims)

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/xarray/core/dataarray.py in compute(self, **kwargs)
    891         """
    892         new = self.copy(deep=False)
--> 893         return new.load(**kwargs)
    894 
    895     def persist(self, **kwargs) -> "DataArray":

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/xarray/core/dataarray.py in load(self, **kwargs)
    865         dask.array.compute
    866         """
--> 867         ds = self._to_temp_dataset().load(**kwargs)
    868         new = self._from_temp_dataset(ds)
    869         self._variable = new._variable

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/xarray/core/dataset.py in load(self, **kwargs)
    739 
    740             # evaluate all the dask arrays simultaneously
--> 741             evaluated_data = da.compute(*lazy_data.values(), **kwargs)
    742 
    743             for k, data in zip(lazy_data, evaluated_data):

~/miniconda3/envs/pangeo/lib/python3.7/site-packages/dask/base.py in compute(*args, **kwargs)
[...]

PR #255, opened 2021-01-18
Add pleiades grid path & better grid_path error message

timothyas:
Address #254 by
1. Introducing a better error message in the case that a filestore's grid_path has not been specified, yet the user wants to `read_grid`
2. Add `grid_path` as argument to `get_dataset` so user can easily provide the grid path on the fly
3. Add what seem to be grid paths for LLC on pleiades

Now if a user has a filestore where `grid_path` is not specified and makes the call:
```
model.get_dataset( etc ..., read_grid=True)
```
 
(note `read_grid`=True is the default) This block of code gives the user something that is hopefully more interpretable...
```
if read_grid and self.store.grid_path is None:
            raise TypeError('Cannot read grid if grid_dir is not specified as argument or in filestore (e.g. llcreader.known_models)')
```

If there is a known grid_path in this case, they can then rerun with:
```
model.get_dataset( etc ..., read_grid=True, grid_path=/path/to/the/goods)
```


timothyas:
@dhruvbalwada would you be able to see if this fixes things for you? I'm getting a strange dask related error if I try to print out any data array, e.g. `print(ds.XC)` or `print(ds.Theta)`. Basically these show up as numpy arrays but xarray is expecting dask arrays. This might be because I hacked together a python setup on Pleiades pretty hastily.

<details>
  <summary>The error output is shown here</summary>

```
Traceback (most recent call last):
  File "test_xmitgcm.py", line 22, in <module>
    print(ds.Theta)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/common.py", line 134, in __repr__
    return formatting.array_repr(self)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 484, in array_repr
    summary.append(repr(arr.coords))
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/coordinates.py", line 74, in __repr__
    return formatting.coords_repr(self)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 393, in coords_repr
    coords, title="Coordinates", summarizer=summarize_coord, col_width=col_width
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 373, in _mapping_repr
    summary += [summarizer(k, v, col_width) for k, v in mapping.items()]
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 373, in <listcomp>
    summary += [summarizer(k, v, col_width) for k, v in mapping.items()]
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 327, in summarize_coord
    return summarize_variable(name, var.variable, col_width, marker)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 292, in summarize_variable
    values_str = inline_variable_array_repr(var, values_width)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 262, in inline_variable_array_repr
    return inline_dask_repr(var.data)
  File "/home1/tasmit12/.local/lib/python3.6/site-packages/xarray/core/formatting.py", line 232, in inline_dask_repr
    assert isinstance(array, dask_array_type), array
```  
</details>

rabernat:
Thanks for working on this Tim!

dhruvbalwada:
Yeah, I can give it a run. Can you tell me how to install the right PR for xmitgcm?

timothyas:
If you have xmitgcm cloned from github then you can:
```
git remote add timothyas https://github.com/timothyas/xmitgcm.git
git fetch timothyas
git checkout --track timothyas/fix_pleiades_grid_path
```

timothyas:
Just confirming that I tested all of this on Sverdrup, which also uses a local filestore and had no problems - everything is read in as a dask array and the additions from this PR work smoothly. Not sure about what's going on on Pleiades.

dhruvbalwada:
I am having to rebuild part of my environment because I had just installed jlab3, and realized that i can't get it to work (authentication doesn't seem to work), so now downgrading. Will get back in a bit.

timothyas:
Sounds good, thanks @dhruvbalwada! I added an Access Token to use git on Pleiades since I have it set up with html there and ssh was a pain. It wasn't required but github emailed me saying it will be soon.

dhruvbalwada:
Does this seem like the correct output? 

Screen Shot 2021-01-18 at 2 45 25 PM

timothyas:
Looks good to me!

timothyas:
Ok I think it's good to go then. I'll work on a test, then remove the draft status. Coming soon ...

xmitgcm/llcreader/llcmodel.py
```diff
@@ -787,7 +787,8 @@ def _check_iters(self, iters):
 
     def get_dataset(self, varnames=None, iter_start=None, iter_stop=None,
                     iter_step=None, iters=None, k_levels=None, k_chunksize=1,
-                    type='faces', read_grid=True, grid_vars_to_coords=True):
+                    type='faces', read_grid=True, grid_vars_to_coords=True,
+                    grid_path=None):
```
rabernat, line 791: Why would we need to be able to override `grid_path` here? What not just create a different Model object instead?

I would favor removing this option from `get_dataset`.

xmitgcm/llcreader/llcmodel.py
```diff
@@ -817,6 +818,8 @@ def get_dataset(self, varnames=None, iter_start=None, iter_stop=None,
             Whether to read the grid info
         grid_vars_to_coords : bool, optional
             Whether to promote grid variables to coordinate status
+        grid_path : str, optional
+            override or provide llcmodel's filestore grid_path
```
rabernat, line 822: remove

xmitgcm/llcreader/llcmodel.py
```diff
@@ -881,6 +884,13 @@ def _if_not_none(a, b):
 
         # grid stuff
         read_grid = read_grid and len(self.grid_varnames)!=0
+        if read_grid and grid_path is not None:
+            if self.store.grid_path is not None:
+                warnings.warn(f'Overriding store.grid_path={self.store.grid_path} with: {grid_path}')
+            self.store.grid_path = grid_path
+
+        if read_grid and self.store.grid_path is None:
+            raise TypeError('Cannot read grid if grid_dir is not specified as argument or in filestore (e.g. llcreader.known_models)')
```
rabernat, line 893: Just seems like unnecessary complexity.

rabernat (changes requested):
Thanks a lot for this @timothyas!

It would be great if we could have tests that only run on Pleaides. That would allow us to test the local Model classes just like we do with the remote ones:
https://github.com/MITgcm/xmitgcm/blob/ea0a18e1ef2470669e381ae9c51e2c0b052071e9/xmitgcm/test/test_llcreader.py#L161-L168

But perhaps we leave that for a future PR.

xmitgcm/llcreader/llcmodel.py
timothyas: Sure thing, I just though it would be nice but it's not essential. I'll get rid of it

rabernat: One confusing thing about our API is that you can specify all of these parameters twice.

I prefer to think that the `Model` initialization specifies what is actually stored on disk, while the `get_dataset` call explains how you want it presented as an Xarray dataset. Going forward, I would be in favor of refactoring the API to remove this duplication.

xmitgcm/llcreader/llcmodel.py
timothyas: Done

xmitgcm/llcreader/llcmodel.py
timothyas: Just the assignment, right? I removed that, but kept the error message (although removed the "as argument" part).

xmitgcm/llcreader/llcmodel.py
timothyas: Done

timothyas: Whoops just seeing your comment about refactoring, didn't mean to prematurely close out. I agree!

rabernat:
LGTM! Take it off draft mode?

timothyas:
I was going to take a stab at a test while I'm at it, one sec

xmitgcm/test/test_llcreader.py
```diff
@@ -120,9 +120,15 @@ def test_vector_mate_error(local_llc90_store, varname):
...
+        try:
+            return llcreader.PleiadesLLC2160Model()
+        except OSError:
+            return llcreader.ECCOPortalLLC2160Model()
     else:
-        return llcreader.ECCOPortalLLC4320Model()
+        try:
+            return llcreader.PleiadesLLC4320Model()
+        except OSError:
+            return llcreader.ECCOPortalLLC4320Model()
```
timothyas, line 132: Ok here's a simple change for tests, basically if on pleiades, get a Pleiades model, otherwise get the portal version

xmitgcm/test/test_llcreader.py
```diff
@@ -187,7 +193,10 @@ def test_ecco_portal_latlon(ecco_portal_model):
 ########### ASTE Portal Tests ##################################################
 @pytest.fixture(scope='module')
 def aste_portal_model():
-    return llcreader.CRIOSPortalASTE270Model()
+    try:
+        return llcreader.SverdrupASTE270Model()
+    except OSError:
+        return llcreader.CRIOSPortalASTE270Model()
```
timothyas, line 199: Same thing on Sverdrup, seemed like a quick way to get it done...

xmitgcm/test/test_llcreader.py
rabernat: I love that you are trying to get this covered by tests. But I don't think this is the right code pattern to use. We don't want the logic of whether tor run the tests dictated by try / except, because unrelated errors could also raise these exceptions. Instead, use a pytest parametrization or parametrized fixture to accomplish this, plus a flag `pytest.mark.skip` if we are not on the appropriate supercomputer.

timothyas: That makes sense. I changed it now to have a separate module for these tests, `test_llcreader_local.py` (name isn't perfect, because there are still local tests in test_llcreader ... but open to suggestions). Now this has the skipif logic to handle whether we're on the super computer, and someone can easily run only these tests by themselves with `py.test test_llcreader_local.py`. Just tried on Sverdrup and Pleiades, and the appropriate tests were skipped. I'm curious what you think.

I see this makes it more explicit about when the portal vs "supercomputer" version is being run, the downside is that I basically copied and pasted the same portal tests and made slight changes to grab pleides/sverdrup versions. This seems not ideal, and there are essentially double the tests to maintain, but I couldn't think of a better way to do it (quickly...).

rabernat: Sorry for letting this hang for a while, and thanks for the work you have put in.

This still isn't quite what I had in mind, and I apologize for not being more clear. I agree with your assessment that duplicating the tests is not the right way to go.

The better way to go is to use pytest fixtures and parametrization. We already have a parametrized fixture for the model.

https://github.com/MITgcm/xmitgcm/blob/a4beec22e4e57150f1362758f6190b6d5ecf62cb/xmitgcm/test/test_llcreader.py#L120-L125

This automatically triggers parametrization to all of the tests that depend on this fixture. We can add more test cases by adding to `param` and then modifying the fixture code accordingly. We want to expand this fixture to test the local models as well as the portal ones (probably renaming it in the process, since it is no longer specific to the ECCO portal).

To handle missing data, we have a check _inside_ this fixture to see if we are on the appropriate server. If not, we call `pytest.skip(reason="we are not on pleaides")`, and the test suite moves on.

Does that make sense?

timothyas: Hey @rabernat, thanks for the patient explanation. I think I'm on the same page now, and this definitely makes more sense, not duplicating tests etc.

xmitgcm/test/test_llcreader.py
```diff
@@ -117,18 +118,29 @@ def test_vector_mate_error(local_llc90_store, varname):
...
-        return llcreader.ECCOPortalLLC4320Model()
 
-def test_ecco_portal_faces(ecco_portal_model):
+        if not os.path.exists('/home6/dmenemen'):
+            pytest.skip("Not on Pleiades")
+        else:
+            if request.param[1]==2160:
+                return llcreader.PleiadesLLC2160Model()
+            else:
+                return llcreader.PleiadesLLC4320Model()
```
timothyas, line 137: The new (and improved?) fixture, for pleiades. Lots of if/else's, but I think this gets the job done efficiently.

rabernat: Yes, I always end up using that sort of pattern in my fixtures too. I agree it's not super pretty, but it works and avoids the repetition of the previous approach.

timothyas: Awesome. Thanks for the guidance @rabernat. Something is up with the tests, my hunch is that it's the ecco portal since I'm getting similar errors on my master branch which is even with upstream/mater.

I can re-trigger these on another day and flag you when it looks good.

timothyas:
Hey @rabernat, looks like the ECCO portal is back on its feet and we're getting the holy green check mark. Let me know if you think anything else needs to be changed or if it's good to go - thanks!

PR #262, opened 2021-05-04
Q: Add exorad flux variable to variables.py

AaronDavidSchneider:
Hi,

this PR is more of a question. 

We are currently developing a package (called `exorad`) for MITgcm which we plan to publish open source (merge into MITgcm) in the future.

`xmitgcm` is such a useful package when it comes to opening MITgcm output. However (as far as I can see), its restricted to the variables declared in `available_diagnostics.log` and `variables.py`.

Therefore, my question: What would be the best practice? 
- Tell the users of our package to extend their installation of `xmitgcm`
- Merge variable names to `xmitgcm` (e.g. with such a PR) even though our package is yet not public
- add the option for the user to add specific variable names

I apologize if the last option is already implemented and I oversaw it.

### DIFF
--- a/xmitgcm/variables.py
+++ b/xmitgcm/variables.py
@@ -757,7 +757,12 @@ package_state_variables = {
         standard_name="ADJvice",
         long_name='dJ/dvice: Sensitivity to meridional ice drift',
         mate='ADJuice',
-        units='dJ/(m s-1)'))
+        units='dJ/(m s-1)')),
+    'EXOBFlux': dict(dims=['k_p1','j', 'i'], attrs=dict(
+        standard_name="EXOBFlux",
+        long_name='bolometric radiative flux from exorad',
+        mate='EXOBFlux',
+        units='W/m2')),
 }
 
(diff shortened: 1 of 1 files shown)


timothyas:
Hey @AaronDavidSchneider 👋 As far as I can tell those are the proper ways to get xmitgcm to read in variables from an MITgcm run. I can only mention what I do: when I have added new variables I either add them to `variables.py` in my own branch, or I go through the trouble to make new diagnostics via the MITgcm (so they show up in available_diagnostics.log). In my opinion, the former is better for quick/scratch work, while the latter is preferred for things you feel confident won't change.

If you're interested in adding variables to the MITgcm diagnostics, so that you can manage them with `data.diagnostics`, etc then I could give some comments to help with that (but maybe raise a separate issue on MITgcm/MITgcm for that one).

AaronDavidSchneider:
Thanks for your feedback!
I already tried to use the diagnostics package for the output but then messed it up at some point :D

I will close this PR then and might reopen it when our package is eventually released.

Thanks!

Issue #278, opened 2021-08-17
LLCreader2160 Sea Ice Velocity 


LizzWebb:
Hi all, 

I am trying to plot the U component of Sea Ice Velocity from the ECCO Portal for model 2160. To keep it simple, I am choosing time =0. However, the plot does not look reasonable. For example, it indicates sea ice velocity throughout, including at the equator, and very little velocity in the arctic (face 6). I am wondering if I am making an error in implementation? My lines of code are below, and I have attached the plot. 

Thank you in advance! 

```
model = llcreader.ECCOPortalLLC2160Model()
ds_SIvel= model.get_dataset(varnames=['SIuice','SIvice'], k_levels=[0])
SeaIce = ds_SIvel.SIuice
SeaIce.isel(time=0).plot.pcolormesh(cmap='coolwarm',vmin=-1, vmax=1, col='face', col_wrap=4)
```

![SIuice_world](https://user-images.githubusercontent.com/83596765/129655289-20cc435f-f5d4-492b-82de-99680a9effb1.png)

mjlosch:
@LizzWebb 
The model (at the time of simulation, defaults have changed by now) has non-zero ice velocities also where there is no ice. They are just free drift velocities of ice with zero thickness. That's also why the Arctic velocities are so much smaller (because that's where the ice is, which does not move in free drift). But there's seems to be a problem with llcreader for this, because the drift field does not look physical.

I can reproduce that just for the Arctic with something like
```
from xmitgcm import llcreader
model = llcreader.ECCOPortalLLC2160Model()
ds=model.get_dataset(k_chunksize=90)
ds.SIuice.isel(time=100,face=6).plot.pcolormesh(ax=ax[0,0],vmin=-1,vmax=1)
```
and similarly for SIvice. Plotting SIheff (which lives on different grid points) works fine:
```
ds.SIheff.isel(time=100,face=6).plot.pcolormesh(ax=ax[0,0],vmin=0,vmax=4)
```
Interestingly the ocean velocities look fine, too (they use the same grid points as SIuice/vice).

Is this a bug in llcreader, or in the data set? @rabernat @menemenlis ?

Maybe related, I found this:
https://github.com/MITgcm/xmitgcm/blob/3b341ad8f43ea2b82e408c7c85011514deddfde6/xmitgcm/llcreader/llcutils.py#L11
Shouldn't that be 
```
    points = ['C', 'W', 'S']
```
to use/load all three types of masks?

[...]

Issue #191, opened 2020-02-24
ecco portal weird looking wind stress


apatlpo:
I am loading wind stress data from the ecco portal but the data looks weird:

![image](https://user-images.githubusercontent.com/11750960/75198883-84f88d00-5716-11ea-92a8-9b9713d430b4.png)

Has anybody seen this before?

See the following [gist](https://gist.github.com/apatlpo/1f2bd9c106c3de980b012666c84c69bb) for more details about how I got the figure above.

ping @rabernat

rabernat:
So my first impression about this is that it is a mask issue. The way the data are compressed is by removing all of the land points, as defined in hFacC / hFacW / hFacS. To uncompress, we need to fill in all of these missing points. The nature of how this image looks distorted makes me think that there is a mismatch between the mask we are assuming and the true mask used to compress the data.

Right now, the mask is determined based on the grid location:
https://github.com/MITgcm/xmitgcm/blob/876786c413488f6be0cb0b479064a9cd2f68b6b5/xmitgcm/llcreader/llcmodel.py#L35-L47

Since `oceTAUX` is at the U velocity point (`i_g, j`), it uses the same mask as U (`hFacW`). You can confirm that U does indeed look right--this is how I debugged the routines initially. But for some reason, this is not working for oceTAUX, oceTAUY.

It would be good if we could verify with the creators of this dataset (i.e. Dimitris), whether a different mask was used in compressing the wind stress.

apatlpo:
thanks Ryan, this makes very much sense.
While waiting for an answer from Dimitris, I'll tweak xmitgcm to use other masks and see if it helps.

rabernat:
One clue is that taux and tauy seem to have the exact same file size as the c-grid variables:
```
-r--r--r-- 1 dmenemen g26209  140637812 Mar  4  2017 oceSflux.0000092240.data.shrunk
-r--r--r-- 1 dmenemen g26209  140637812 Mar  4  2017 oceTAUX.0000092240.data.shrunk
-r--r--r-- 1 dmenemen g26209  140637812 Mar  4  2017 oceTAUY.0000092240.data.shrunk
```

That suggests they should have the same mask. So try to force it to use the `hFacC` mask.

rabernat:
Yes, that is it. This works
```diff
diff --git a/xmitgcm/llcreader/llcmodel.py b/xmitgcm/llcreader/llcmodel.py
index 8b5e069..90a3aee 100644
[...]

PR #279, opened 2021-08-17
fix masking for SIuice and SIvice in the same way as for oceTAUX/Y

mjlosch:
This fixes issue #278 as PR #192 fixed issue #191, by add SIu/vice to mask_override.


rabernat:
Amazing, thank you Martin! 💪

PR #280, opened 2021-08-17
Use correct mask suffixes

rabernat:
This apparent typo was noted by @mlosch in #278.

It looks severe. However, how could things have possibly worked without it. This really makes me wish we had better test coverage for the llcreader stuff. 🤦

### DIFF
--- a/xmitgcm/llcreader/llcutils.py
+++ b/xmitgcm/llcreader/llcutils.py
@@ -8,7 +8,7 @@ def load_masks_from_mds(grid_dir):
     import xmitgcm
     ds = xmitgcm.open_mdsdataset(grid_dir, iters=None, geometry='llc')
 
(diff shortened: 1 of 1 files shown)


mjlosch:
@rabernat I apologise for my ignorance, but where would you think this makes a difference? In my simple test (read and plot llc data with `model = llcreader.ECCOPortalLLC2160Model()`) I don't see any difference with or without this fix. How could I test it?

rabernat:
I think this code may have been used to _generate_ the zarr masks that are used to decompress the LLC data. Indeed it is only called in one place--`copy_masks_from_mds_to_zarr`--and that is not called anywhere else in xmitgcm. So basically this is orphaned code

We know those masks are correct, since they work. So this is really orphaned and untested code. I'll merge.

PR #285, opened 2021-09-27
ENH: 2D chunks for all

jklymak:
This allows 2D chunking for all mds files, not just the LLC cases.  

This is a total hack at this point/proof of concept.  However, it works for me, and allows me to get 2-D chunks and parallel work on a machine that doesn't have enough memory for my model run (for some reason I can't discern - each variable is only 5 Gb).  

- [ ] needs docs
- [ ] needs tests
- [ ] probably needs to be properly interested with the other chunking options.

```python
with xm.open_mdsdataset(data_dir, prefix=['means'], 
                                endian="<", geometry='cartesian',
                                chunks="2D") as ds:
   print(ds)
```

Here Nz=400:
```
...
Data variables:
    VVEL     (time, Z, YG, XC) float64 dask.array<chunksize=(1, 1, 1216, 1440), meta=np.ndarray>
    UVEL     (time, Z, YC, XG) float64 dask.array<chunksize=(1, 1, 1216, 1440), meta=np.ndarray>
    THETA    (time, Z, YC, XC) float64 dask.array<chunksize=(1, 1, 1216, 1440), meta=np.ndarray>
...
```


mjlosch:
@jklymak I realise this is old, but would it be possible to contribute a short test for this new functionality?

mjlosch:
@IvanaEscobar The tests all passed for python 3.11 to 3.13 on my local machine. Not sure what happened here, looks like it has to do with the CI itself.

jklymak:
@mjlosch possibly this summer, but not very soon!  I'll move to draft, or anyone is more than welcome to take this over.

IvanaEscobar:
@mjlosch the error is addressed in PR #358 . The CI needed updated gfortran compilers in each environment tested.

PR #293, opened 2022-01-04
Edit README to update version requirement for xarray

fluidnumerics-joe:
* Relevant to https://github.com/MITgcm/xmitgcm/issues/292

PR #295, opened 2022-01-06
Ignore deprecation warnings

fraserwg:
Updated `test_ecco_portal_iterations` and `test_aste_portal_iterations` so that they don't fail when deprecation warnings are raised.  Currently xarray and distutils are raising DeprecationWarnings we have no control over causing these tests to fails. See also pull request #293


rabernat:
Amazing, thanks!

Issue #303, opened 2022-05-09
NaNs in data set for large number of files.


mjlosch:
I am trying to read 310 files with 6 3D `llc90` fields each (i.e. [50,13,90,90] grid points per 3D field) with 
```
ds = open_mdsdataset('../mydirectory',prefix=['diags3D'],delta_t=deltat,ref_date=refdate,geometry='llc')
```
This works fine until I increase the number files `diags3D` to larger than 186 (not sure why this number). When I do that the variables with `ds.isel(time=range(186))` contain only `nan`, but the variables with `ds.isel(time=range(186,310))` are OK. E.g. `dst0.iter[185].values` gives `nan`, but `dst0.iter[186].values` gives the correct "iteration" number, same for 3D fields. `ds.time` is correct for all records. 

Any idea what's going on?

mjlosch:
User error. What a surprise! The first 186 files `diags3D.*` contained 6 fields, the remaining 124 files contained 7 fields. 

Is there a way to make open_mdsdataset (actually `xarray.combine_by_coords`) ignore the extra variable in such a case? 
I tried the flag `data_vars=minimal`, but that didn't help.
My solution to this is now to post-process all 186 files to add a dummy field with zeros before I can use open_mdsdataset, but that's really not optimal.

rabernat:
Hi Martin and sorry for the slow reply here!

> I tried the flag `data_vars=minimal`, but that didn't help.

That would have been my best guess. 😞  Too bad it didn't work.

My recommendation for a quick workaround would be to open the files as two different datasets (specifying `iters` manually in `open_mfsdataset`), harmonize them however you prefer, and then manually concatenate them with xarray.

mjlosch:
Thanks Ryan, I think it should be possible to fix this (so that only the extra fields have NaN and not everything). I tried to combine just two dataset with different fields (read with open_mdsdataset) with `xarray.combine_by_coords`, and there it works as expected (somehow: the extra field is NaN where it was not available, but all other fields are OK). When I read these two directly with open_mdsdata, then I can reproduce the problem. I am not sure where this happens.

mjlosch:
Just for the record: When I try the above combination (`compat='no_conflicts', coords='minimal'`) on my entire data set I get this error after some waiting time:
```
Traceback (most recent call last):
  File "/home/ollie/mlosch/MITgcm/xmitgcm/xmitgcm/mds_store.py", line 202, in open_mdsdataset
    iternum = int(iters)
[...]

PR #304, opened 2022-05-17
drop data variables not common to all datasets before combining them

mjlosch:
This fixes the problem described in #303: if in a series of output files from the diagnostics package some files contain more fields than others. Currently the `xr.combine_by_coords(..., combat='override', ...)` sets all fields to NaNs in files that contain fewer fields. 

Clearly, changing the number of fields in a diagnostics-output stream is not ideal, but this fix would have saved me from debugging my notebooks for a few days.


xmitgcm/mds_store.py
```diff
@@ -239,6 +239,20 @@ def open_mdsdataset(data_dir, grid_dir=None,
                 datasets = [open_mdsdataset(
                         data_dir, iters=iternum, read_grid=False, **kwargs)
                     for iternum in iters]
+                # drop all data variables not common to the datasets,
+                # this should be done by xr.combine_by_coords, but
+                # with the "overide" option, it does not work properly
+                all_data_vars = set(datasets[0].data_vars)
```
timothyas, line 245: This line will fail if `datasets == []`, which can happen and then the user will probably be staring at a weird error.

timothyas:
Hey @mjlosch, this looks generally good to me, but I made a detailed comment above. Since I mentioned that I would help add a test, I've made fixed that and added a test - would you like me to push those changes? The push would also contain a merge with upstream master, so we should actually see the tests pass.

xmitgcm/mds_store.py
mjlosch: thanks for the comment. We could catch this case and only do lines 242 to 255 `if datasets != []`

timothyas: That is what I was thinking as well. Do you want me to push those changes to your branch?

mjlosch: Yes, please, go ahead

timothyas:
Ah, good to see the green check mark. I made a test that copies the LLC90 meta/data pair to a new iteration and hides "ETAN" from the new meta file. I originally made this as the last test in the `test_mds_store.py` module, and when I did this, the following would pass: `pytest test_mds_store -k test_drop_uncommon_diagnostics`  but running the whole test module `pytest test_mds_store` would fail. I have no idea why, but by moving this test earlier in the sequence, everything works just fine.

mjlosch:
looks great! At one point you'll have to explain to me the subtleties of using brackets or not.

I would have included all of the new code in the new if statement, but this looks very good, too.

Just a note: when I run the test locally I get a lot of "DeprecationWarning"s, but the do not have anything to do with the modifications of the PR.

I agree, it's ready to go.

timothyas:
Great, thanks @mjlosch. I think I used brackets just to get a one liner, but it's not really necessary. I'm seeing the deprecation warnings too, it's on the todo list for the next plumbing update.

I think this has little chance of breaking anyone else's code, so unless there's any rebuttal I will merge this soon to push it through.

Issue #307, opened 2022-06-24
Surface velocities from ECCO data portal


cspencerjones:
For a future project, I wanted to use surface velocities from the ECCO data portal (to avoid egress charges). They are available as part of the regional data released [here](https://data.nas.nasa.gov/ecco/data.php?dir=/eccodata/llc_4320/regions/global/SSU). The data is "shrunk" so it probably makes sense to use existing llcreader code to unshrink it.

I downloaded a single time slice of SSU and SSV and so far my code looks like this:
```
from xmitgcm import llcreader
from fsspec.implementations.local import LocalFileSystem
fs = LocalFileSystem()
store = llcreader.BaseStore(fs, base_path='/swot/SUM01/spencer/LLCsurf')
model = llcreader.LLC4320Model(store)
ds_faces = model.get_dataset(varnames=['SSU'])
```

However, xmitgcm isn't aware that SSU is a possible variable name, so I get an error:
```
KeyError                                  Traceback (most recent call last)
<ipython-input-7-86109449e317> in <module>
----> 1 ds_faces = model.get_dataset(varnames=['SSU'])

~/.conda/envs/pangeo1/lib/python3.9/site-packages/xmitgcm/llcreader/llcmodel.py in get_dataset(self, varnames, iter_start, iter_stop, iter_step, iters, k_levels, k_chunksize, type, read_grid, grid_vars_to_coords)
    897 
    898         # get the data in facet form
--> 899         data_facets = {vname:
    900                        self._get_facet_data(vname, iters, k_levels, k_chunksize)
    901                        for vname in varnames}

~/.conda/envs/pangeo1/lib/python3.9/site-packages/xmitgcm/llcreader/llcmodel.py in <dictcomp>(.0)
    898         # get the data in facet form
    899         data_facets = {vname:
--> 900                        self._get_facet_data(vname, iters, k_levels, k_chunksize)
    901                        for vname in varnames}
    902 

~/.conda/envs/pangeo1/lib/python3.9/site-packages/xmitgcm/llcreader/llcmodel.py in _get_facet_data(self, varname, iters, klevels, k_chunksize)
    741     def _get_facet_data(self, varname, iters, klevels, k_chunksize):
    742         # needs facets to be outer index of nested lists
--> 743         dims = _VAR_METADATA[varname]['dims']
    744 
    745         if len(dims)==2:

KeyError: 'SSU'
```

[...]

PR #205, opened 2020-06-08
modified open_mdsdataset to add extra_variables

cspencerjones:
Fixes #185 . The user will be able to specify extra variables that are non-standard and don't exist in available_diagnostics.log.

PR #308, opened 2022-06-27
Custom Grid Vars & Extra Variables for llcreader

timothyas:
This implements two things:

1. Within the core `mds_store`: allow a user to add "on-the-fly" grid variables, or really any meta/data pair that does not have a 10 digit iteration number. For example, if we have a file `smooth2Dnorm001.meta/data` in our run_directory:

```python
custom_grid_vars = {'smooth3Dnorm001': {
    'dims': ['k', 'j', 'i'], 
    'attrs': {
        'standard_name': 'normalization_factor',
        'long_name': 'WC01 normalization factor',
        }
    }
}
ds = open_mdsdataset(run_dir, grid_dir, ..., custom_grid_variables=custom_grid_vars)
```

2. For `llcreader`, add this capability and #205 to add extra_variable meta information to an llcmodel, e.g.

```python
fs = LocalFileSystem()
store = llcreader.BaseStore(fs, base_path=data_dir, grid_path=grid_dir)
model = llcreader.LLC90Model(store)
ds = model.get_dataset(varnames=['smooth3Dnorm001'], extra_variables=custom_grid_vars, ...)
```

@cspencerjones: Hopefully with this PR my comment to #307 makes more sense. You could use this to add your SSU/SSV variables on-the-fly as with the extra_variables capability for mds_store. The only issue is if those SSU/SSV files are on a different grid, things will get more complicated, but if they're on the global grid this should work no problem.

I'll keep this as a draft for now because I need to add tests.

### DIFF
--- a/xmitgcm/test/test_xmitgcm_common.py
+++ b/xmitgcm/test/test_xmitgcm_common.py
@@ -278,7 +278,7 @@ def file_md5_checksum(fname):
 
 # find the tar archive in the test directory
 # http://stackoverflow.com/questions/29627341/pytest-where-to-store-expected-data
-@pytest.fixture(scope='module', params=_experiments.keys())
+@pytest.fixture(scope='function', params=_experiments.keys())
 def all_mds_datadirs(tmpdir_factory, request):
     return setup_mds_dir(tmpdir_factory, request, _experiments)
 


timothyas:
Tests currently fail due to the issue second described in #312. They do pass when the pytest fixture scope is set to "function", as described in that issue.

Issue #309, opened 2022-07-13
`open_mdsdataset` fails if more than 999 vertical levels


fraserwg:
# The problem
If using 1,000 vertical levels or more (admittedly this may be a very niche use case) then `open_mdsdataset` fails.

# The cause
If the number of levels is 1,000 or more instead of writing the number of levels, the MITgcm code writes `***` in the `Levs` column of `available_diagnostics.log`. See the diagnostics 11 to 13 below for an example
```
 Total Nb of available Diagnostics: ndiagt=   225
------------------------------------------------------------------------------------
  Num  |<-Name->|Levs|  mate |<- code ->|<--  Units   -->|<- Tile (max=80c)         
------------------------------------------------------------------------------------
     1 |SDIAG1  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #1
     2 |SDIAG2  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #2
     3 |SDIAG3  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #3
     4 |SDIAG4  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #4
     5 |SDIAG5  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #5
     6 |SDIAG6  |  1 |       |SM      L1|user-defined    |User-Defined   Surface   Diagnostic  #6
     7 |SDIAG7  |  1 |       |SU      L1|user-defined    |User-Defined U.pt Surface Diagnostic #7
     8 |SDIAG8  |  1 |       |SV      L1|user-defined    |User-Defined V.pt Surface Diagnostic #8
     9 |SDIAG9  |  1 |    10 |UU      L1|user-defined    |User-Defined U.vector Surface Diag.  #9
    10 |SDIAG10 |  1 |     9 |VV      L1|user-defined    |User-Defined V.vector Surface Diag. #10
    11 |UDIAG1  |*** |       |SM      MR|user-defined    |User-Defined Model-Level Diagnostic  #1
    12 |UDIAG2  |*** |       |SM      MR|user-defined    |User-Defined Model-Level Diagnostic  #2
    13 |UDIAG3  |*** |       |SMR     MR|user-defined    |User-Defined Model-Level Diagnostic  #3
```
[...]

PR #310, opened 2022-07-13
added fix for reading mds data with a large number of vertical levels.

fraserwg:
This resolves issue #309.

### DIFF
--- a/xmitgcm/test/test_utils.py
+++ b/xmitgcm/test/test_utils.py
@@ -946,6 +946,24 @@ def test_parse_diagnostics(all_mds_datadirs, layers_mds_datadirs):
         assert ad[key] == val
 
 
+def test_parse_diagnostics_bad_levels(tmpdir):
+    from xmitgcm.utils import parse_available_diagnostics
+    # Try parsing a 'bad' available_diagnostics.log file
+    bad_diags_path = tmpdir / 'available_diagnostics.log'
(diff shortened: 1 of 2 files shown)


timothyas:
Hey @fraserwg, sorry this has been sitting around with no attention for a while. I'm curious, what happens when you read in a dataset in this case? Does `xmitgcm` properly figure out how many levels there are?

fraserwg:
So when `xmitgcm` parses the available diagnostics file, this fix just sets the number of levels to `np.NaN` if there's 1,000 or more vertical levels for the given diagnostic. The only place this variable seems to be subsequently used is when using the layers package. There's already a warning in place for if the number of layers cannot be inferred and I think this should trigger that warning as required. Short of refactoring the `MITgcm` code which writes the `available_diagnostics.log` file  I don't think there's any other way to infer the number of levels available.

timothyas:
Ah, I see now, thanks for that explanation. This does only seem to impact the layers package, but I agree with how you wrote it. That is, I think it's better to catch the error and have `levs` set before going into the subsequent `if/else` tree so that the code is flexible for when `levs==1` or in the case you're catching when `np.isnan(levs)`.

Test looks good to me too. I say LGTM, and will merge early tomorrow unless there is any disagreement on that.

timothyas:
Thanks for your efforts @fraserwg!

PR #311, opened 2022-07-31
Update usage.rst

yangleir:
Change 'print(ds_llc['S']dims)' to 'print(ds_llc['S'].dims)'


timothyas:
Thanks for this fix @yangleir!

PR #313, opened 2022-08-15
Single llc face

IvanaEscobar:
using face4 to read facets instead of face0


IvanaEscobar:
Meant to Pr this to my own fork...

PR #342, opened 2025-03-06
ENH: allow cache clearing from open_mdsdataset

jklymak:
I had to chase the caching logic down in the source code, so this just allows us to reset the cache when we use mds_opendataset.  Default is to _not_ clear the cache.


mjlosch:
Thanks @jklymak, this looks like a very useful feature (to me). Do you think, you can add a test or modify an existing test to actually use this feature (I am guessing in `xmitgcm/test/test_mds_store.py`)?

PR #344, opened 2025-04-07
MNT: add flag for decode_cf

jklymak:
Without this, `xarray` emits a bunch of warnings.


jklymak:
@mjlosch - for sure we can add to this PR to address other issues.  If you have solutions, always feel free to push to my PRs or supersede them with your own if that is cleaner.  I can probably look at this in a day or two.  Thanks!

PR #348, opened 2025-04-24
One face region

IvanaEscobar:
Reviving #335 
---
Old patches used for applications of domains contained in a single face. 

Helps to:
- set `offset_vars` when `file_metadata` only has a single variable
- use `find_concat_dim_facet` if DataArray input doesn't contain a `face` dimension
- use llc regions that aren't starting on `facet0`, reads `nz` not the first facet which works in existing domains defined in fxn `get_extra_metadata`


IvanaEscobar:
This PR needs to be fixed to work with tests!

PR #357, opened 2025-11-07
Repair CI

IvanaEscobar:
Updating `zarr` library version to fix CI errors. 

Error only occurred in Github Actions miniconda environments, based linux architectures. I could **not** reproduce errors in local M2 MacOS generated environments. 

CI now passes again.

### DIFF
--- a/ci/environment-py3.12.yml
+++ b/ci/environment-py3.12.yml
@@ -8,7 +8,7 @@ dependencies:
   - numpy
   - pytest
   - future
-  - zarr >=2.15
+  - zarr >=3.1.3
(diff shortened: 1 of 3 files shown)


IvanaEscobar:
I would suggest a merge of this PR before too much time passes. 

I'm not completely aware of changes done on each Python library used in xmitgcm, nor any changes done to miniconda used by GitHub actions. The cadence of updates to those dependencies can be addressed detangled from user's PRs. It seems that letting PRs sit too long introduces errors in CI that are irrelevant to the PRs themselves.

